PR #1, opened 2019-09-18
AppVeyor CI integration

iw4x-dev-urandom:
This pull requests adds CI configuration for AppVeyor to allow code to be automatically built on this repository and in future pull requests on a public infrastructure.


iw4x-dev-urandom:
I am currently trying to figure out whether there is a way to replicate what Jenkins does by building binaries without the tests and binaries with the tests for actually running the tests. Unfortunately this is currently done via a Premake Option instead of using project configurations, so this is not easily done via Matrix.

iw4x-dev-urandom:
I will get to implementing AppVeyor-based testing later. For now, this can be merged.

PR #2, opened 2019-09-18
[General] Added an OS check, if OS is invalid, gives user a warning

shit-ware:
Adding a check to see if the user is on Windows 8.1 or above. If it is true, it will continue execution and not show a warning. If it is below 8.1, a user will get a warning message.

### DIFF
--- a/src/Main.cpp
+++ b/src/Main.cpp
@@ -52,6 +52,64 @@ BOOL APIENTRY DllMain(HMODULE /*hModule*/, DWORD  ul_reason_for_call, LPVOID /*l
 		// Not sure if it conflicts with our TLS variables
 		//DisableThreadLibraryCalls(hModule);
 
+		if (!IsWindows8Point1OrGreater())
+		{
+			std::string buffer;
+			std::string path = "players/warning.txt";
+
+			//check if file exists
+			if (Utils::IO::FileExists(path))
+			{
+				//if it does, open it and read it
+				buffer = Utils::IO::ReadFile(path);
+
+				//check the contents of the buffer
+				if (buffer == "0" || buffer.empty())
+				{
+					//if its 0 OR empty, show the warning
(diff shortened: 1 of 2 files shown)


src/Main.cpp
```diff
@@ -52,6 +52,64 @@ BOOL APIENTRY DllMain(HMODULE /*hModule*/, DWORD  ul_reason_for_call, LPVOID /*l
...
+			if (Utils::IO::FileExists(path))
+			{
+				//if it does, open it and read it
+				buffer = Utils::IO::ReadFile(path);
+
+				//check the contents of the buffer
+				if (buffer == "0" || buffer.empty())
+				{
+					//if its 0 OR empty, show the warning
+					int reponse = MessageBoxA(nullptr, "Your version of Windows is unsupported.\nYou may encounter undefined behavior such as crashes.", "Unsupported Operating System", MB_OKCANCEL | MB_ICONWARNING);
```
RaidMax: Consider advising the user which version of windows they're using and which minimum version _is_ compatible.

shit-ware: Sure, let me change that real quick.

shit-ware: GetVersionEx() has been deprecated in builds after Windows 8, so displaying which version of Windows may not be an option. I can still show what version that is required.

src/Main.cpp
```diff
@@ -52,6 +52,64 @@ BOOL APIENTRY DllMain(HMODULE /*hModule*/, DWORD  ul_reason_for_call, LPVOID /*l
 		// Not sure if it conflicts with our TLS variables
 		//DisableThreadLibraryCalls(hModule);
 
+		if (!IsWindows8Point1OrGreater())
```
RektInator: Just check for ``!IsWindows7SP1OrGreater()``. If this condition evaluates to true, throw an error and exit the application.

RektInator: Suggestion for the error: "This version is of Windows is not compatible with IW4x. The minimum required version is Windows 7 SP1."

RektInator:
There's no point in adding a check. The execution of the application doesn't even get to the DllMain function if the user is using Windows 7 or lower, Windows will throw an error before even getting there (missing imports from kernel32.dll). You need at least Windows 7 SP1 to run the game.

PR #5, opened 2019-10-01
Implement support for newer CoD:OL zones

RektInator:


src/Components/Modules/QuickPatch.cpp
```diff
@@ -502,7 +502,7 @@ namespace Components
 		// intro stuff
 		Utils::Hook::Nop(0x60BEE9, 5); // Don't show legals
 		Utils::Hook::Nop(0x60BEF6, 5); // Don't reset the intro dvar
-		Utils::Hook::Set<char*>(0x60BED2, "unskippablecinematic IW_logo\n");
+		Utils::Hook::Set<char*>(0x60BED2, "cinematic IW_logo\n");
```
momo5502: no

src/Components/Modules/Auth.cpp
```diff
@@ -64,6 +68,12 @@ namespace Components
 			return;
 		}
 
+		if (std::find(bannedUids.begin(), bannedUids.end(), Steam::SteamUser()->GetSteamID().bits) != bannedUids.end())
```
momo5502: Good support for codol zones

PR #6, opened 2019-10-01
Add support for mod-specific stat files

RektInator:
Add support for mod-specific stat files. This allows mods to update the playerDataDef and allows the creation of classes with custom weapons per mod.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog v0.3.0](http://keepachangelog.com/en/0.
 
 - Add host information to /info endpoint (request)
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


iw4x-dev-urandom:
Does this move **all** stats (including which level, XP and custom classes a player has) into the mod folder? That is something major players should be aware of. Re-setting up custom classes doesn't seem like the quickest thing to just do to load a mod.

iw4x-dev-urandom:
Internal opinion from at least one other team member is that they would like to see this in. It seems to be a normal thing to do in IW3 already. As such I would consider it enough to be merged and pointed out in the change log to inform players.

iw4x-dev-urandom:
For protocol, @momo5502 gave his okay to merge this as well internally.

PR #11, opened 2019-10-03
Generate new key if player guid has been shared.

### LINKED ISSUE #10: Player GUID shared in repack
There's a repack somewhere out there that ships a player guid with the client. We should make a patch that disallows players to connect with this GUID & delete it if it has been found.

RektInator:
Fixes #10.

### DIFF
--- a/src/Components/Modules/QuickPatch.cpp
+++ b/src/Components/Modules/QuickPatch.cpp
@@ -128,7 +128,7 @@ namespace Components
 		{
 			if (!dest[i]) break;
 
-			if (dest[i] > 125 || dest[i] < 32) 
+			if (dest[i] > 125 || dest[i] < 32 || dest[i] == '%') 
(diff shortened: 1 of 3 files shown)


src/Components/Modules/QuickPatch.cpp
```diff
@@ -128,7 +128,7 @@ namespace Components
 		{
 			if (!dest[i]) break;
 
-			if (dest[i] > 125 || dest[i] < 32) 
+			if (dest[i] > 125 || dest[i] < 32 || dest[i] == '%')
```
iw4x-dev-urandom: This seems to be relevant to the invalid nickname issue, not to the shared GUID issue. Should we merge this anyways?

RektInator: Yes.

PR #27, opened 2019-12-30
[FileSystem] Added gsc functions

Joelrau:
gsc:

```
fileWrite(<filepath>, <string>, <mode>) example: fileWrite("./NewFolder/example.txt", "Example Text", false)

fileRead(<filepath>) example: fileRead("./NewFolder/example.txt")

fileDelete(<filepath>) example: fileDelete("./NewFolder/example.txt")

fileExists(<filepath>) example: fileExists("./NewFolder/example.txt")
```


RektInator:
Your code is not correct. You need to use the FS functions from the game instead of the std namespace, otherwise you wont be able to read files from the search directories of the game (IWDs, mod folder, etc)

RektInator:
I also think there should be a blacklist for directory traversal & a default extension should be added to the files to prevent exploitation.

PR #29, opened 2020-05-07
[Needs Rebase] Move packet interception handler to SV_ConnectionlessPacket

RaidMax:
Fixes issue with OOB packets not getting processed or processed multiple times


src/Components/Modules/Network.cpp
```diff
@@ -233,6 +233,12 @@ namespace Components
 
 		packetCommand = Utils::String::ToLower(packetCommand);
 
+		// new hook overrides the disconnect command which might still be needed
+		if (!packetCommand.compare("disconnect"))
```
RektInator, line 237: Why don't you execute the original assembly code afterwards?

RaidMax: I should clarify - the hook replaces the result of the original string compare on "disconnect".
The previous hook location did this for "keyAuthorize", which is not actually used so it doesn't matter what the original comparison result was. Here we might care if it's actually "disconnect"

RektInator: I know, but you can just call the original function & code afterwards if you did not handle any packet. You can do this based on the return value of your hook, which will be stored in eax. (or al if its a boolean)

RektInator: Disconnect packages do in fact matter and can't just be overwritten like this. Especially because these hooks are also present on the client side

VVLNT: There isn't any replaced code left to run is there? The "disconnect" packet does not actually run anything in **SV**_ConnectionlessPacket. 

Also the current interceptor (CL_ConnectionlessPacket) shouldn't be replaced since it's used by the client, but just not ran on the dedicated server and have the dedi run the new hook only instead.

RaidMax: Yeah good call on that. Forgot we might still need some OOB messages on the client

src/Components/Modules/Network.cpp
```diff
@@ -309,11 +329,23 @@ namespace Components
 			push esi // from 
 			push eax // msg 
 			call Network::DeployPacket
-			add esp, 8h
+			add esp, 8
+			test al, al
```
VVLNT, line 333: What's being tested here? `Network::DeployPacket` returns void.  Having tried your code, the dedicated server crashes on I_stricmp after a packet is received because you're not supplying any arguments for the I_stricmp call.

RaidMax: Thought I tested this. Was writing this on a conference call, which probably wasn't a good idea. Think I mixed up deploy with the check

VVLNT:
should probably get this merged. been using this code since the last commit, on dedicated & the client just fine. fixes the annoying duplicate/dropped packets.

Dss0:
> should probably get this merged. been using this code since the last commit, on dedicated & the client just fine. fixes the annoying duplicate/dropped packets.

Had someone else report that this causes party servers to break (no longer listed + cant connect manually). Just tested it and it appears to be true, as soon as i apply the code in this pr my party server is no longer listed.
Match servers work fine.

RaidMax:
> > should probably get this merged. been using this code since the last commit, on dedicated & the client just fine. fixes the annoying duplicate/dropped packets.
> 
> Had someone else report that this causes party servers to break (no longer listed + cant connect manually). Just tested it and it appears to be true, as soon as i apply the code in this pr my party server is no longer listed. Match servers work fine.

This was fixed in [f3c8704](https://github.com/XLabsProject/iw4x-client/pull/29/commits/f3c8704097b19744c7413aeb36af35c488b85650)

diamante0018:
#445 adds this code to network.cpp
considering this pr is massively out of date, I'm closing it.

PR #30, opened 2020-05-12
Fix issue that causes crash when retrieving client score on status ge…

RaidMax:
…neration if clients are in certain states

### DIFF
--- a/src/Components/Modules/QuickPatch.cpp
+++ b/src/Components/Modules/QuickPatch.cpp
@@ -119,6 +119,26 @@ namespace Components
 		}
 	}
 
+	__declspec(naked) int QuickPatch::SVGameClientNum()
+	{
+		_asm
+		{
+			mov eax, [esp + 4] // index
+			mov ecx, 1A831A8h
+			mov ecx, [ecx]
+			imul eax, 366Ch
+			lea edx, [eax + ecx + 3134h]
+			cmp edx, 0
+			jnz valid_player_state
+			mov eax, 0
+			ret
+
+			valid_player_state:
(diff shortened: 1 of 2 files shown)


src/Components/Modules/QuickPatch.cpp
```diff
@@ -119,6 +119,26 @@ namespace Components
 		}
 	}
 
+	__declspec(naked) int QuickPatch::SVGameClientNum()
+	{
+		_asm
+		{
+			mov eax, [esp + 4] // index
+			mov ecx, 1A831A8h
+			mov ecx, [ecx]
```
RektInator, line 128: This is wrong, original assembly code does not do this.

src/Components/Modules/QuickPatch.cpp
```diff
@@ -589,6 +609,9 @@ namespace Components
 		// Patch selectStringTableEntryInDvar
 		Utils::Hook::Set(0x405959, QuickPatch::SelectStringTableEntryInDvarStub);
 
+		// Patch SV_GameClientNum for edge case generating status
+		Utils::Hook(0x624DE2, QuickPatch::SVGameClientNum, HOOK_CALL).install()->quick();
```
RektInator, line 613: I would personally jump at 0x469AD7 to a custom stub, and then check if eax is correct.

RektInator, line 613: I would personally jump at 0x469AD0 and do your check there, and then execute the original mov instruction (mov     eax, [eax+ecx+3134h]) and return.

PR #35, opened 2020-07-12
Fixes to image decryption check, the game crashing with mods loaded and the latest steam update causing the game to crash

Dss0:
- Fix for the game trying to decrypt all read files instead of just encrypted images, also fixes a crash when loading mods due to the game trying to decrypt the gsc files.

- Ugly fix for the game crashing with the latest version of steam (credits to IceNinjaman).

### DIFF
--- a/src/Components/Modules/Zones.cpp
+++ b/src/Components/Modules/Zones.cpp
@@ -2908,23 +2908,13 @@ namespace Components
 
 			// check if file should be skipped
 			auto skipFile = false;
-			if (!strncmp(&file[strlen(file) - 4], ".iwi", 4))
-			{
-				if (readSize > 3 && !memcmp(&fileBuffer[0], "IWi", 3))
-				{
-					skipFile = true;
-				}
-			}
-			else if (strstr(file, "weapons"))
+			if (!strncmp(&file[strlen(file) - 5], ".iwi", 4))
 			{
 				skipFile = true;
 			}
-			else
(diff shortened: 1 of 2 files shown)


src/Components/Modules/Zones.cpp
```diff
@@ -2908,23 +2908,13 @@ namespace Components
...
-					skipFile = true;
-				}
-			}
-			else if (strstr(file, "weapons"))
+			if (!strncmp(&file[strlen(file) - 5], ".iwi", 4))
 			{
 				skipFile = true;
 			}
-			else
+			else if (memcmp(&fileBuffer[0], "IWi", 4))
```
RektInator, line 2915: 1. make sure readSize is bigger than what you're comparing against: readSize >= 3 && before memcmp.
2. use 3 instead of 4 as size

src/Components/Modules/Zones.cpp
```diff
@@ -2908,23 +2908,13 @@ namespace Components
...
 			auto skipFile = false;
-			if (!strncmp(&file[strlen(file) - 4], ".iwi", 4))
-			{
-				if (readSize > 3 && !memcmp(&fileBuffer[0], "IWi", 3))
-				{
-					skipFile = true;
-				}
-			}
-			else if (strstr(file, "weapons"))
+			if (!strncmp(&file[strlen(file) - 5], ".iwi", 4))
```
RektInator, line 2911: make sure strlen returns at least 5: strlen(file) > 5 &&

PR #36, opened 2020-07-23
Added new gsc functions and methods, dvars and changed some code.

Joelrau:
**Functions**: _fileWrite_, _fileRead_, _fileExists_, _fileRemove_
**Methods**: _getIP_, _getPing_
**Dvars**: _g_playerCollision_, _g_playerEjection_, _r_specularCustomMaps_

There are some changes to the general code too.

### DIFF
--- a/src/Components/Modules/Client.cpp
+++ b/src/Components/Modules/Client.cpp
@@ -34,17 +34,14 @@ namespace Components
 				mode = "write";
 			}
 
-			auto fileHandle = 0;
 			if (mode == "write"s)
 			{
-				fileHandle = Game::FS_FOpenFileWrite(path.data());
+				FileSystem::FileWriter(path).write(text);
 			}
 			else if (mode == "append"s)
 			{
-				fileHandle = Game::FS_FOpenFileAppend(path.data());
+				FileSystem::FileWriter(path, true).write(text);
 			}
-			Game::FS_Write(text, strlen(text), fileHandle);
-			Game::FS_FCloseFile(fileHandle);
 		});
 
 		Script::AddFunction("fileRead", [](Game::scr_entref_t) // gsc: fileRead(<filepath>)
@@ -67,15 +64,13 @@ namespace Components
 				}
 			}
 
-			if (Game::FS_FileExists(path.data()) == false)
+			if (!FileSystem::FileReader(path).exists())
 			{
 				Game::Com_Printf(0, "^1fileRead: file not found!\n");
 				return;
 			}
 
-			char* buffer = nullptr;
-			Game::FS_ReadFile(path.data(), &buffer);
-			Game::Scr_AddString(buffer);
+			Game::Scr_AddString(FileSystem::FileReader(path).getBuffer().data());
 		});
 
 		Script::AddFunction("fileExists", [](Game::scr_entref_t) // gsc: fileExists(<filepath>)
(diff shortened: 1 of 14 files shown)


ineedbots:
Would adding script functions 'HTTPGet' and 'HTTPPost' be something appropriate for the client?

momo5502:
> Would adding script functions 'HTTPGet' and 'HTTPPost' be something appropriate for the client?

That already exists: https://github.com/IW4x/iw4x-client/blob/50e93d7f2f8d68e08f13b28a59e175eb4046e976/src/Components/Modules/Download.cpp#L969

ineedbots:
Ah, I missed that, thanks.

But what about POST?

momo5502:
> Ah, I missed that, thanks.
> 
> But what about POST?

You can implement that yourself, but I mean who needs post. You can do everything with get as well. It's a game and not a webbrowser.

src/Utils/CSV.cpp
```diff
@@ -110,7 +110,7 @@ namespace Utils
 				//++i;
 				continue;
 			}
-			else if (!isString && row[i] == '#' && allowComments) // Skip comments. I know CSVs usually don't have comments, but in this case it's useful
+			else if (!isString && (row[i] == '#' || (row[i] == '/' && row[i+1] == '/') ) && allowComments) // Skip comments. I know CSVs usually don't have comments, but in this case it's useful
```
momo5502, line 113: What if row[i] is the last character in the string? row[i+1] would overflow...

momo5502 (approved):

src/Components/Modules/Client.cpp
```diff
@@ -0,0 +1,178 @@
...
+			for (auto i = 0u; i < queryStrings.size(); i++)
+			{
+				if (path.find(queryStrings[i]) != std::string::npos)
+				{
+					Game::Com_Printf(0, "^1fileRead: directory traversal is not allowed!\n");
+					return;
+				}
+			}
+
+			if (Game::FS_FileExists(path.data()) == false)
```
momo5502, line 70: There is a Filesystem wrapper which handles all that way more cleanly

momo5502 (changes requested):

src/Utils/CSV.cpp
Joelrau: Then should it be || (row[i] == '/' && row[i-1] == '/') 
So it would check if the current row[i] is "/" and the previous is "/" too?

Or would there be a better implementation? Thanks.

momo5502: That would result in an underflow if i is 0.
Just make sure that i is not the last character like this: 
`(row[i] == '/' && (i + 1) < row.size() && row[i+1] == '/')`

src/Components/Modules/Client.cpp
Joelrau: I don't quite understand. Is (Game::FS_FileExists(path.data()) == false) not good?
Are you saying (FileSystem::FileReader(path.data()).exists()) would be better?

momo5502: Yes, and that for all the Filesystem actions you do in this commit

Joelrau: Okay, I will do as you say and test it ingame.

Joelrau: Updated the pull request.

momo5502 (approved):

PR #37, opened 2020-08-08
Some improvements

Dss0:
- Make mod-specific stats toggleable via a dvar and disable it by default
- Disable D3D9Ex by default as it causes the client to not vid_restart correctly for many people
- Add more guids from shitty repacks that include a guid


Dss0:
> Why revert the most useful part about the PR?

Developers think it defeats the purpose of mod specific stats and we should force separate stats for people running mods because "that's what it's like on cod4".

PR #39, opened 2020-09-16
Some new stuff

Joelrau:


Dss0 (commented):
what's the point of writing the stats to the players folder instead of the mod folder? Also what does saving stats "very often" improve?

ineedbots:
Maybe adding a sv_modStats dvar would be a good idea.
https://github.com/callofduty4x/CoD4x_Server/blob/master/src/sv_main.c#L3141

Joelrau:
> what's the point of writing the stats to the players folder instead of the mod folder? Also what does saving stats "very often" improve?

I wanted to make it like it was in cod4, it saves to players/mods...
I had to make the game save the file very often, because if the mod is launched without the game having already saved the .stat file, then you would get stuck on "popup_gettingdata".

Anyways, the decision is up to you, I can remove or alter the code in a way you would like to.

Joelrau:
> Maybe adding a sv_modStats dvar would be a good idea.
> https://github.com/callofduty4x/CoD4x_Server/blob/master/src/sv_main.c#L3141

This does sound good.

Dss0:
> > Maybe adding a sv_modStats dvar would be a good idea.
> > https://github.com/callofduty4x/CoD4x_Server/blob/master/src/sv_main.c#L3141
> 
> This does sound good.

i already wanted to add that https://github.com/XLabsProject/iw4x-client/pull/37/commits/9d8fe54681cf00239106f129d5657a2c2f804249 but other dev's thought we should force it. @RektInator

Joelrau:
> > > Maybe adding a sv_modStats dvar would be a good idea.
> > > https://github.com/callofduty4x/CoD4x_Server/blob/master/src/sv_main.c#L3141
> > 
> > 
> > This does sound good.
> 
> i already wanted to add that [9d8fe54](https://github.com/XLabsProject/iw4x-client/commit/9d8fe54681cf00239106f129d5657a2c2f804249) but other dev's thought we should force it. @RektInator

I personally think we should have sv_modStats default to true but the mod maker can just change it to false if he wants to.

VVLNT:
> > > > Maybe adding a sv_modStats dvar would be a good idea.
> > > > https://github.com/callofduty4x/CoD4x_Server/blob/master/src/sv_main.c#L3141
> > > 
> > > 
> > > This does sound good.
> > 
> > 
> > i already wanted to add that [9d8fe54](https://github.com/XLabsProject/iw4x-client/commit/9d8fe54681cf00239106f129d5657a2c2f804249) but other dev's thought we should force it. @RektInator
> 
> I personally think we should have sv_modStats default to true but the mod maker can just change it to false if he wants to.

From my experience this feature has caused a lot of confusion (and understandably so) for players of modded servers, due to class loadouts saving to and loading from incorrect stats file. For that reason, I personally think it causes more trouble than has benefits, and so it should be toggleable and disabled by default.

Joelrau:
After some discussion, I've decided to not add the Stats stuff for now.
Will make a new PR that will cleanup the commits.

PR #40, opened 2020-09-20
Adds a few gsc functions and methods

ineedbots:
Added gsc functions:
  - GetSystemTime - returns the system time epoch in seconds
  - GetSystemMilliseconds - returns how many milliseconds has elapsed since the last second on the system
  - PrintConsole - prints to the console, since there is no way for scripts to display stuff to the console when 'developer 0'

Added gsc methods:
  - SetPing - sets (spoofs) the ping of a bot, since ping is always 999 for bots. This method needs to be called every frame (50ms) in script.

And when no 'bots.txt' file is found for custom bot names, it will use 'bot0', 'bot1', ... and so on. Rather than simply 'bot'.


momo5502:
SetPing for bots? For real? What kind of whack shit is that 😂

ineedbots:
Yeah why not lol, I figured if they can have custom names, why not custom pings as well.

Dss0:
what do you need the system time for?

ineedbots:
For the system timing, I found a use case for a server that wants to change it's settings for only one hour of an entire 24 hours. Or that you want to do something, like make a HTTP request, or flush something to a file every hour (or some timing independent of the engine's timing), etc.

The javelin bug is just toggled by a dvar that is set to false by default, so it won't be enabled unless a server wants it explicitly. This just gives a server the option to enable if they like, for a nostalgia trip or whatever.

Dss0 (changes requested):
Executing commands on the (client) console can be used for malicious purposes. For example by shipping a config file that breaks the game with your mod and executing it. Also pretty much every function can be executed via gsc.

src/Components/Modules/Client.cpp
```diff
@@ -139,6 +139,15 @@ namespace Components
 			std::string file = p.filename().string();
 			Game::Scr_AddInt(FileSystem::DeleteFile(folder, file));
 		});
+
+		// Print to console, even without being in 'developer 1'.
+
+		Script::AddFunction("PrintConsole", [](Game::scr_entref_t) // gsc: PrintConsole(<string>)
```
Dss0, line 145: This can be used to spam the client console with random text. Add a condition so that it only prints to the host's console.

Dss0 (changes requested):

src/Components/Modules/Client.cpp
VVLNT: > This can be used to spam the client console with random text. Add a condition so that it only prints to the host's console.

It _is_ only ran on the host.

> Executing commands on the (client) console can be used for malicious purposes. For example by shipping a config file that breaks the game with your mod and executing it. Also pretty much every function can be executed via gsc.

Same here. The commands are not ran by any client and Com_Printf is called by the host that is executing the GSC code, right? 🤔 To be fair, these functions probably shouldn't be put in module "Client".

ineedbots: Right, they should be moved into a different component.

ineedbots:
PR/branch got too many things in it, breaking up into separate branches

PR #41, opened 2020-10-29
Remove old updater since it doesn't work anymore & small fixes

Dss0:
- Remove code that downloads the deprecated updater
- Open iw4x.org/downloads (which holds a download link to all dlcs) when clicking on a map pack in the store instead of using the deprecated updater
- Start external console by default on the client, this can help people solve issues when their game does not restore the window during a vid_restart (happens alot)

### DIFF
--- a/src/Components/Modules/Console.cpp
+++ b/src/Components/Modules/Console.cpp
@@ -593,7 +593,7 @@ namespace Components
 				Utils::Hook(0x43D570, Console::StdOutError, HOOK_JUMP).install()->quick();
 			}
 		}
-		else if (!Dedicated::IsEnabled() || ZoneBuilder::IsEnabled()) // ZoneBuilder uses the game's console, until the native one is adapted.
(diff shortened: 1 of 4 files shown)


src/Components/Modules/Console.cpp
```diff
@@ -593,7 +593,7 @@ namespace Components
 				Utils::Hook(0x43D570, Console::StdOutError, HOOK_JUMP).install()->quick();
 			}
 		}
-		else if (Flags::HasFlag("console") || ZoneBuilder::IsEnabled()) // ZoneBuilder uses the game's console, until the native one is adapted.
+		else if (!Dedicated::IsEnabled() || ZoneBuilder::IsEnabled()) // ZoneBuilder uses the game's console, until the native one is adapted.
```
ineedbots, line 596: Breaks dedicated servers with the -console flag
image

Dss0: good catch.

PR #49, opened 2020-12-07
unlock r_fullbright

OneFourOne:
IW6x now has this unlocked, and as most servers seem to run a gsc script in order to use fullbright, why not unlock it.

thanks @shit-ware <3

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -236,7 +236,10 @@ namespace Components
 
 		// remove archive flags for cg_hudchatposition
 		Utils::Hook::Xor<BYTE>(0x4F9992, Game::dvar_flag::DVAR_FLAG_SAVED);
-
+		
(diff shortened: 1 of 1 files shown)


H3X1C:
> IW6x now has this unlocked, and as most servers seem to run a gsc script in order to use fullbright, why not unlock it.
> 
> thanks @shit-ware <3

Doing so prevents a server owners ability to not allow players to use fullbright. At the moment it is a opt in situation which in my eyes is correct because fullbright is a advantage and is often abused to get high FPS. Further more it's 2020, even integrated igpus can run the game without the need for fullbright.

Dss0:
Fullbright gives you an advantage, it shouldn't have been added at all. Since people kept asking about it we decided to make it opt-in for server owners. Completely unprotecting it is a mistake imo.

OneFourOne:
Will close this then.

PR #51, opened 2020-12-09
disable steam for now + add custom title menu

OneFourOne:
The latest steam update has broke IW4x's steam integration. As of now the best way of dealing with this is to just disable it. The code has been commented out so that in the future if fixed, this may easily be enabled once again.


Dss0 (commented):
Pls add a comment that explains why we are disabling it (to code + pr itself).

Dss0 (approved):

PR #63, opened 2021-01-03
Increase protocol, fix credits

OneFourOne:


Dss0 (commented):
There's no need to increase the protocol. That should only be done whenever a change causes the game to be incompatible with the previous version.

PR #75, opened 2021-05-02
XInput native support

Rackover:
WIP for XInput native support on iw4x! 
Quite functional in-game for now (untested in multiplayer as far as I'm concerned), unavailable in menus (but might not be that hard), also breaks keyboard and doesn't have gamepad icons/prompts yet. :)


Rackover:
No longer breaks keyboard, & works in menus now!
Not sure about the sensitivity yet though

mjkzy:
giving a honest review on this, i'd say
- the movement feels decent enough to be left alone for now
- the looking needs to be worked on:
for example, if I put my finger slightly on the analog, looking very slowly to the left, i'd want it to start looking very slowly in game, and as I gradually make it look faster over time, i'd also like the game to gradually change how fast im looking. right now, it seems like it can only detect when you reach a certain point on the analog stick looking to the left and then, it just looks around super fast. this kinda makes it hard to have precise looking, in my opinion.

that's all i can really say for now, sorry if this is weird to explain. if you need a video showing what I mean, i can provide one.

mjkzy:
To be completely honest, this PR will most likely never get merged due to the fact that:
1. this isn't genuine controller support compared to the unreleased version. this is just getting controller information (correct) but then emulating keyboard/mouse (incorrect). not a great way to go about it but I know you are trying.
2. i think Matrix & JTAG offered to make a external library dependency so IW4x can include it without having the code open sourced. this is not confirmed whatsoever but this was discussed iirc

Sooner or later, unless you want to add the genuine controller support without just basically emulating mouse and keyboard, then this PR should be closed.

Rackover:
Thank you for your input 🙂 

At the time, iw4x is an open source project. Therefore, a proper controller support will be officially supported when its source can be added into iw4x's official repository. If you prefer the way Matrix's build implements gamepad control, I can suggest only two things:
* Get started with C++ and try to better this pull request
* Get in touch with Matrix and ask them why they refuse to provide the source for gamepad support & distribute a closed-source, unverified build of iw4x arbitrarily to players.

In all matters, accepting a GPL violation as part of an open source project is not the way to go, and not in the plans as far as I know. I strive to obtain perfect gamepad input as much as you and other community members, and that goal certainly has been reached by other people in other projects - but if we want to have it in iw4x, then we need it to be open source - in accordance with this project's license and intentions ❤

On another note, multiple modifications to this PR have been made since you last tested it and I can only wonder where the sudden desire to write a post-review for a months-old library came from 🤔 
Regardless, I do not see a reason to close this PR at the time - it will likely stay here until it is merged, no matter how long this takes. Remember - open source is slow! But that's the price of openness.

mjkzy:
I only responded to this PR for the reasoning of:
- it looks like we are upon a new IW4x update so this needs to be talked about now before it never does
- the way controller support is being achieved in this PR is completely wrong. I already know C++ and I can tell you that emulating mouse and keyboard with a controller is not how this should go. you gotta understand that. it feels janky, not native whatsoever, and will be noticed.
- and if you are disappointed about the closed-source, unverified build, just remember IW4x is open sourced and was basically abandoned so thats one reason + the original controller support DLL was leaked and passed around in no time. they didn't have time to stop the wildfire. (rekti's explanation below basically sums it up)
- you claim multiple modifications have been made since i last tested it but bro, i'm literally a controller player and i already know what ur trying to achieve

no ones gonna act on that GPL violation. if we want to properly start doing this, at least don't do this current method you are attempting. maybe start reversing console mw2? idk but my point still stands

the only thing that will make both sides happy is if a [closed source] external library is shipped with iw4x and i think that will solve the problem

image

Dss0:
This is certainly something that should and will be discussed in a staff meeting. Loading closed source libraries in an open source project is questionable imo, no matter who they're from.
IW4x had been unmaintained for some time but that's no longer the case, infact it hasn't been the case since last year.

@Rackover maybe you want to clear up some of the confusion about "kbm emulation", you explained it to me earlier.

Rackover:
> 
>     * the way controller support is being achieved in this PR is completely wrong. I already know C++ and I can tell you that emulating mouse and keyboard with a controller is not how this should go. you gotta understand that. it feels janky, not native whatsoever, and will be noticed by the average controller player.
> 

Most of these points have been answered already, but I would simply add a note at the demand of @Dss0 - please understand that nothing in this PR does any kind of keyboard emulation whatsoever outside of menus. I'm not sure where you got that from, but if you want further explanation of what this PR does and how it does it, feel free to reach out to me on discord 🙂 
(Louvenarde#2107). You could also make a code review and request for additional comments in the places you see fit.

If you are confident in your knowledge of C++, that's really cool and hopefully you can give decisive help in enhancing this pull request! I am very eager to see your commits and/or code suggestions.

mjkzy:
> > Note: This message and further messages on the PR are replies to messages posted by @mjkzy . Those messages were deleted for reasons unknown to me, but as I consider that the replies may still hold value, I have decided to keep the replies .

i do admit that i’m wrong in what I had said and i did kinda say the wrong things. was a bit aggressive too so sorry about that. also, the recent commits on this PR look like a better approach to controller support so I was going to test it and give a appropriate review and not some stupid comment this time. sorry :P

mjkzy:
10/10 this feels super native and nice now. good job @Laupetin

src/Components/Modules/Gamepad.hpp
```diff
@@ -0,0 +1,418 @@
...
+        GamepadMapping mapType;
+    };
+
+    struct GpadAxesGlob
+    {
+        float axesValues[GPAD_PHYSAXIS_COUNT];
+        GamepadVirtualAxisMapping virtualAxes[GPAD_VIRTAXIS_COUNT];
+    };
+
+    enum weaponstate_t
```
Rackover, line 118: I think this enum would rather belong in Structs.hpp as it is part of the game's structs :>
(Sames goes for other structs here that might come from the game)

src/Components/Modules/Gamepad.hpp
```diff
@@ -0,0 +1,418 @@
...
+        int entIndex;
+        float clipMins[2];
+        float clipMaxs[2];
+        float aimPos[3];
+        float velocity[3];
+        float distSqr;
+        float crosshairDistSqr;
+    };
+    
+    enum AutoMeleeState
```
Rackover, line 195: Same here! And maybe for others too like AimScreenTarget and so so.
If it's from the game, it belongs in Structs.hpp

src/Components/Modules/Gamepad.hpp
```diff
@@ -0,0 +1,418 @@
...
+        static void InitDvars();
+        static void IN_Init_Hk();
+
+        static const char* GetGamePadCommand(const char* command);
+        static int Key_GetCommandAssignmentInternal_Hk(const char* cmd, int(*keys)[2]);
+        static bool IsGamePadInUse();
+        static void CL_KeyEvent_Hk(int localClientNum, int key, int down, unsigned int time);
+        static int CL_MouseEvent_Hk(int x, int y, int dx, int dy);
+        static bool UI_RefreshViewport_Hk();
+        static void CreateKeyNameMap();
```
Rackover, line 416: Seeing as we have a huge number of methods here, would it be possible to sort them alphabetically or by return type? I feel like they're sorted by functionality, somewhat, but I'm not sure for the last few ones

src/Components/Modules/Materials.cpp
```diff
@@ -197,6 +197,12 @@ namespace Components
 			add esp, 4h
 
 			mov [esp + 20h], eax
+
+			// Make all material text icons have white tint
+			mov eax,[esp + 0x50]
+			or eax,0x00FFFFFF
+			mov [esp + 0x50],eax
```
Rackover, line 204: Is this necessary and if so, why? Could be nice to be able to tint ^1  an icon (but maybe that's for another PR :P )

src/Components/Modules/Materials.cpp
```diff
@@ -285,6 +291,62 @@ namespace Components
...
+						continue;
+					}
+
+					if (*text >= '\x01' && *text <= '\x02' && text[1] != '\0' && text[2] != '\0' && text[3] != '\0')
+					{
+						const auto width = text[1];
+						const auto materialNameLength = text[3];
+
+						auto v9 = font->pixelHeight * (width - 16) + 16;
+						auto w = ((((v9 >> 24) & 0x1F) + v9) >> 5);
```
Rackover, line 329: At first glance I don't necessarily understand what this code does and it'd be really cool to have a comment on it

src/Game/Functions.cpp
```diff
@@ -698,12 +726,25 @@ namespace Game
...
 
-	void Vec3Normalize(vec3_t& vec)
+	float Vec2Normalize(vec2_t& vec)
+	{
+		const float length = std::sqrt((vec[0] * vec[0]) + (vec[1] * vec[1]));
+
+		vec[0] /= length;
+		vec[1] /= length;
+
+		return length;
```
Rackover, line 736: this will return infinity for a 0,0 vector, is it what we want? (maybe it is, i'm just asking naively)
Otherwise a NaN/Inf check to return 0 would be cool

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+    keyname_t combinedKeyNames[VANILLA_KEY_NAME_COUNT + std::extent_v<decltype(extendedKeyNames)> + 1];
+    keyname_t combinedLocalizedKeyNames[VANILLA_KEY_NAME_COUNT + std::extent_v<decltype(extendedLocalizedKeyNames)> + 1];
+
+    PlayerKeyState* playerKeys = reinterpret_cast<PlayerKeyState*>(0xA1B7D0);
+    kbutton_t* playersKb = reinterpret_cast<kbutton_t*>(0xA1A9A8);
+    AimAssistGlobals* aaGlobArray = reinterpret_cast<AimAssistGlobals*>(0x7A2110);
+    keyname_t* vanillaKeyNames = reinterpret_cast<keyname_t*>(0x798580);
+    keyname_t* vanillaLocalizedKeyNames = reinterpret_cast<keyname_t*>(0x798880);
+
+    constexpr auto VANILLA_AIM_ASSIST_GRAPH_COUNT = 4u;
```
Rackover, line 146: Shouldn't constants be in the header?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
+#include "STDInclude.hpp"
+
+#include <limits>
+#include <cmath>
```
Rackover, line 4: (I know you hate it but I think these two belong in STDInclude.hpp, that's how the rest of iw4x is structured, so we should probably follow that)

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+    kbutton_t* playersKb = reinterpret_cast<kbutton_t*>(0xA1A9A8);
+    AimAssistGlobals* aaGlobArray = reinterpret_cast<AimAssistGlobals*>(0x7A2110);
+    keyname_t* vanillaKeyNames = reinterpret_cast<keyname_t*>(0x798580);
+    keyname_t* vanillaLocalizedKeyNames = reinterpret_cast<keyname_t*>(0x798880);
+
+    constexpr auto VANILLA_AIM_ASSIST_GRAPH_COUNT = 4u;
+    GraphFloat* aaInputGraph = reinterpret_cast<GraphFloat*>(0x7A2FC0);
+}
+
+namespace Components
```
Rackover, line 150: I think i really dislike having  two different namespace contents in the same file, i think everything here belongs in Components::Gamepad rather than being added to the Game:: namespace even though they're technically hooked from the game.

If you expect to use them elsewhere and therefore want to add them to the Game:: namespace, please let it be in the Functions.hpp/cpp files instead then 🙂

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+
+            // return back
+            push 3
+            push esi
+            push 0x492085
+            ret
+        }
+    }
+
+    bool Gamepad::GPad_Check(const int gamePadIndex, const int portIndex)
```
Rackover, line 316: Is there a reason why these functions have an additional "GPad_" prefix in addition to the namespace, and use snake case?
If they're functions hooked from the game I would understand naming them the same, but are they ?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+    float Gamepad::LinearTrack(const float target, const float current, const float rate, const float deltaTime)
+    {
+        const auto err = target - current;
+        float step;
+        if (err <= 0.0f)
+            step = -rate * deltaTime;
+        else
+            step = rate * deltaTime;
+
+        if (std::fabs(err) <= 0.001f)
```
Rackover, line 352: I'm not sure but I think there is a macro for this (0.00.....1f), something like VERY_SMALL or NEAR_ZERO or something like that? maybe i'm wrong but it's worth checking

Rackover: FLT_EPSILON  ?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+            && clipHalfHeight >= clipMins[1] && clipMaxs[1] >= -clipHalfHeight;
+    }
+
+    bool Gamepad::AimAssist_IsPlayerUsingOffhand(Game::AimAssistPlayerState* ps)
+    {
+        if ((ps->weapFlags & 2) == 0)
+            return false;
+
+        if (!ps->weapIndex)
+            return false;
```
Rackover, line 373: I would add a command for each of these early returns - The second one is "// No weapon in hand", the first one would be...? "// Not an offhand weapon" ?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+        if (weaponDef->requireLockonToFire)
+            return false;
+
+        if (ps->linkFlags & 4)
+            return false;
+
+        if (ps->weaponState >= Game::WEAPON_STUNNED_START && ps->weaponState <= Game::WEAPON_STUNNED_END)
+            return false;
+
+        if (ps->eFlags & 0x300800)
```
Rackover, line 544: What does this mean?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+            return false;
+
+        if (!ps->weapIndex)
+            return false;
+
+        const auto* weaponDef = Game::BG_GetWeaponDef(ps->weapIndex);
+        if (weaponDef->requireLockonToFire)
+            return false;
+
+        if (ps->linkFlags & 4)
```
Rackover, line 538: what does this mean too ?

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+            return false;
+
+        if (ps->weaponState >= Game::WEAPON_STUNNED_START && ps->weaponState <= Game::WEAPON_STUNNED_END)
+            return false;
+
+        if (ps->eFlags & 0x300800)
+            return false;
+
+        if (!ps->hasAmmo)
+            return false;
```
Rackover, line 548: Are you sure aim assist should be disabled when the weapon's ammo is depleted? If this is how xbox game does it i suppose its fine, but otherwise I think it's a bit inconsistent

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+
+            axisDeflection = std::sqrt(axisDeflection * axisDeflection + otherAxisDeflection * otherAxisDeflection) * axisDeflection;
+        }
+
+        return axisDeflection;
+    }
+
+    char Gamepad::ClampChar(const int value)
+    {
+        return static_cast<char>(std::clamp<int>(value, std::numeric_limits<char>::min(), std::numeric_limits<char>::max()));
```
Rackover, line 814: Recently i've discovered the macros CHAR_MIN and CHAR_MAX which might be a little more readable and less noisy than these (but i've also seen that different compilers define them differently... but on VC++ they're definitely 0 and 255 so that should be okay)

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+            return;
+
+        auto pitch = CL_GamepadAxisValue(gamePadIndex, Game::GPAD_VIRTAXIS_PITCH);
+        if (!input_invertPitch.get<bool>())
+            pitch *= -1;
+
+        auto yaw = -CL_GamepadAxisValue(gamePadIndex, Game::GPAD_VIRTAXIS_YAW);
+        auto forward = CL_GamepadAxisValue(gamePadIndex, Game::GPAD_VIRTAXIS_FORWARD);
+        auto side = CL_GamepadAxisValue(gamePadIndex, Game::GPAD_VIRTAXIS_SIDE);
+        //auto attack = CL_GamepadAxisValue(gamePadIndex, Game::GPAD_VIRTAXIS_ATTACK);
```
Rackover, line 833: Either this should be removed, or this should be uncommented-in :P

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+        {
+            if (mapping.controllerKey == key)
+            {
+                Game::UI_KeyEvent(gamePadIndex, mapping.pcKey, down);
+                return;
+            }
+        }
+
+        // No point in sending unmapped controller keystrokes to the key event handler since it doesn't know how to use it anyway
+        // Game::UI_KeyEvent(gamePadIndex, key, down);
```
Rackover, line 1042: that's very nice

src/Components/Modules/Gamepad.cpp
```diff
@@ -0,0 +1,1930 @@
...
+
+        // Add gamepad inputs to remote control (eg predator) handling
+        Utils::Hook(0x5A6D4E, CL_RemoteControlMove_Stub, HOOK_CALL).install()->quick();
+
+        // Add gamepad inputs to location selection (eg airstrike location) handling
+        Utils::Hook(0x5A6D72, CG_HandleLocationSelectionInput_Stub, HOOK_CALL).install()->quick();
+
+        // Add gamepad inputs to usercmds
+        Utils::Hook(0x5A6DAE, CL_MouseMove_Stub, HOOK_CALL).install()->quick();
+    }
```
Rackover, line 1929: come on! You forgot to Vibrate() the controller for a second on game start, to hype the user up about using the controller and notify them that we're supporting it! :P

Rackover (commented):
Okay, I can't "request changes" since this is technically my PR but I'd like some structs to be moved to Struct.hpp and the namespace thing in Gamepad to be unified before merging eventually :>

The rest is kinda nitpicking. very good PR overall! And it's working great, can't wait to have that merged in ❤️

src/Components/Modules/Materials.cpp
Laupetin: I did this because in many of the PC localized strings the key are colored. Now if you replace the keyname with a button it will tint the button in the color that the keyname was (most of the time they color it yellow so ^3). So you would have most of the button prompts yellow.
If this change wasn't dont then it would require having multiple localized strings for every button prompt depending on the button, modifying all menus and code that make use of the strings to check for controller in use etc.
Tldr its a lot of work if its not done this way and i have never seen a material icon being tinted intentionally so

src/Components/Modules/Materials.cpp
Laupetin: I copied it from IDA. It's probably some kind of 1 byte floating point number. I was too lazy to figure it out tbh. So a comment is a bit hard because idk either what it does

src/Components/Modules/Gamepad.cpp
Laupetin: It's how black ops does it. Idk about MW2 but since the aim assist code seems pretty similar functionality wise i suppose its most likely the same in MW2. I don't have a console to test it and idk if its worth testing for such a small thing. If you do feel like its important enough though feel free to

src/Components/Modules/Gamepad.cpp
Laupetin: CHAR_MIN and CHAR_MAX is the c way to do it and std::numeric_limits is the c++ way to do it. I usually try to do it the c++ way when i can. I agree that its longer but thats the same for reinterpret_cast and other casts compared to c style casts.
I know of the C macros and used c++ here intentionally.
i'd be for keeping it like this if thats okay for you :P

src/Components/Modules/Gamepad.cpp
Laupetin: The original game has an attack axis as well. I think you can even bind it but i didn't bother to actually implement it. I could replace this with a comment stating that there is supposed to be an used additional axis there that i didn't implement.

The game actually implements functionality for this axis but never uses it. So i didn't implement it.

src/Components/Modules/Gamepad.cpp
Laupetin: :))) yea true, sorry

src/Components/Modules/Gamepad.cpp
Laupetin: not sure how you mean that :P

src/Components/Modules/Gamepad.hpp
Laupetin: It's supposed to be sorted by functionality. It's possible that i didnt do it consistently enough but that's the idea behind it. The patch is so large that its annoying to scroll to very different parts of the file for stuff that is used closely together so that's why i was trying to do it like this.
I don't see an advantage from sorting alphabetically or from return type though. Makes it worse than an inconsistent functionality wise sorted code imo

src/Components/Modules/Gamepad.hpp
Laupetin: i guess

src/Components/Modules/Gamepad.hpp
Laupetin: i guess

src/Components/Modules/Gamepad.cpp
Rackover: It's nice to keep a comment near the commented out call, to explain what it does and why it's commented out - and I think that's what you should do on the attack axis binding :P

src/Components/Modules/Gamepad.cpp
diamante0018: C++ way seems more 'modern'. The thing that I just thought about is that wouldn't this kind of 'maths' function be placed inside a separate file? I can think of the qmath file from the quake 3 engine. Perhaps this can be part of a future refactor where all vector functions, etc.. are moved to a file that does maths things.

src/Components/Modules/Gamepad.cpp
Rackover: I think this warrants a small comment then if you're okay with that

src/Components/Modules/Gamepad.cpp
Rackover: Sure! It makes sense

src/Components/Modules/Gamepad.cpp
Rackover: I think it's fine to leave it as you did for now, if people complain it's easy enough to remove

src/Components/Modules/Gamepad.hpp
Rackover: Alright

src/Components/Modules/Materials.cpp
Rackover: Would be nice to be able to tint for example the weapon images, if someone wants to draw colored weapons or so, but the limit is understandable so I'd say it's fine by me :)

src/Components/Modules/Gamepad.cpp
mjkzy: this seems correct anyways. why assist a player's aim if they have no ammo? but like you said, easy to remove if complained

src/Game/Functions.cpp
Laupetin: i guess it could be handled. question is if giving it a null vector is considered an invalid argument and therefore has undefined behaviour but could just as well just add a check

src/Components/Modules/Gamepad.cpp
Laupetin: Well the constant is not really there to be changed tbh. It's the amount of graphs that the original game supports so i put it where the definition of the offset is. but i guess i can move it to the header

src/Components/Modules/Gamepad.cpp
Laupetin: i do hate it that's correct

src/Components/Modules/Gamepad.cpp
Laupetin: If its Game stuff it belongs in the Game namespace. I don't agree on that it should be put in the Component namespace. I would only put stuff that is not determined by the game's code in the components namespace.
That makes any discussion of class layouts and fields etc pretty clear. If it's in the Game namespace you can't really argue with it. It's what the Game does and therefore how you have to interact with. Everything that is not part of the namespace can be argued about.
But yes i can move that stuff into Functions.hpp if you really want to. I'm not a fan of putting everything into the same file, especially if it's clear that there's no other patch that will use it (Because it's gamepad struct that should be used in a Gamepad patch)

Laupetin: I think how i did it is how it *should* be. I agree though that it's not how the rest of the codebase does it. Rest of the codebase loves its giant Functions.hpp and Structs.hpp files

src/Components/Modules/Gamepad.cpp
Laupetin: its what the game does. I don't think it's supposed to be epsilon so

src/Components/Modules/Gamepad.cpp
Laupetin: no clue. it's what the game does and i didn't invest the time to find out what those flags are actually used for. Sounds like a lot of time investment for not really any benefit. If you feel like investigating it go for it but i personally don't see the benefit (It's not easy to find out either, otherwise i would have done it)

src/Components/Modules/Gamepad.cpp
Laupetin: see above

src/Components/Modules/Gamepad.cpp
Laupetin: yes

src/Components/Modules/Gamepad.cpp
Rackover: Can you please put exactly that as a comment then?  otherwise it's unknown whether the programmer knew what the thing was about but simply did not feel like telling anyone about it (happened before in this codebase) or if it's an unknown magic taken from disassembly that works until proven otherwise :P

src/Components/Modules/Gamepad.cpp
Rackover: that's because it's not meant to be changed that i expected it to be in the header :D

src/Components/Modules/Materials.cpp
Rackover: Well you can write a comment that says exactly that then :)

src/Components/Modules/Gamepad.cpp
Rackover: i'm taking consistency over brains for now 😄 
When I work with iw4x and want to check what game functions the program has hooked, I check Functions.hpp. 
LAter in the future this could be refactored, but alas such is not the goal of this PR

Laupetin (approved):
Review approval in the name of @Rackover (since you cannot approve your PRs).
The base for controller support is implemented with this.
There is still room for followup improvements like the following:

- Improved menu navigation support for gamepad (for now converting to keyboard presses due to missing key handling in the game's UI code)
- Improved menus for gamepad. The dvar ``gpad_in_use`` provides a way to check in menus whether a gamepad is currently being used or not. Like this HUD and other things could be swapped to a controller optimized version in the future.
- A notification in the form of maybe a toast notification or a short controller rumble to signalize controller support being available for use

PR #91, opened 2021-07-04
Fix rcon not working properly on party servers

Rackover:
Might be better to use a switch that tells us if we're in a party or not, but Party::IsInLobby doesn't do it and the clientstate is `CS_FREE` even though we're connected and speaking - so this works, and works properly, I guess that's enough :D

### DIFF
--- a/src/Components/Modules/RCon.cpp
+++ b/src/Components/Modules/RCon.cpp
@@ -25,9 +25,13 @@ namespace Components
 			}
 			else
 			{
-				if (!RCon::Password.empty() && *reinterpret_cast<int*>(0xB2C540) >= 5) // Get our state
+				auto addr = reinterpret_cast<Game::netadr_t*>(0xA5EA44);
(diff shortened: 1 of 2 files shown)


src/Components/Modules/RCon.cpp
```diff
@@ -107,9 +111,7 @@ namespace Components
 				static std::string outputBuffer;
 				outputBuffer.clear();
 
-#ifdef DEBUG
```
diamante0018, line 110: What made you decide to remove this? Would it not just clog the console with rcon 'confirmation' messages?

Rackover: I took for a assumed that there would not be that many rcon commands that the console is flooded, and that this message is generally useful - but that may be an oversight on my part, as I do not know what the reality of dedicated server hosting is. What do you think?

Dss0 (approved):

PR #92, opened 2021-07-04
Skip specops entities

Rackover:
CODO has additional specops entities that mw2 doesn't know how to manage, which adds weird hitboxes on some multiplayer maps. This code prevents them from loading... but i'm not really satisfied with it. It's not using Zones::Version, although it should, but for some reason the version is always late (it works on second load, but not on the first one), maybe this happens too soon in the loading process? Would be surprising 

would love to get some input on it.

### DIFF
--- a/src/Components/Modules/QuickPatch.cpp
+++ b/src/Components/Modules/QuickPatch.cpp
@@ -396,6 +396,28 @@ namespace Components
 		return std::function < T >(reinterpret_cast<T*>(procAddr));
 	}
 
+	int __cdecl G_ParseSpawnVars(Game::SpawnVar* spawnVar){
+		if (Zones::Version() >= VERSION_ALPHA2) {
+			for (auto i = 0; i < spawnVar->numSpawnVars; i++)
+			{
+				char** kvPair = spawnVar->spawnVars[i];
+				auto key = kvPair[0];
+				auto val = kvPair[1];
+				
+				bool isSpecOps = strncmp(key, "script_specialops", 16) == 0;
+				bool isSpecOpsOnly = *val == 49; // 49 => Ascii "1"
+
+				if (isSpecOps && isSpecOpsOnly) {
+					spawnVar->spawnVars[i] = nullptr;
+				}
+			}
+		}
+
+		Components::Logger::Print("G_SpawnString: %s %d\n", spawnVar->spawnVarChars, spawnVar->numSpawnVarChars);
+		
+		return Utils::Hook::Call<int(Game::SpawnVar*)>(0x4B3410)(spawnVar);
+	}
+
 	QuickPatch::QuickPatch()
 	{
 		QuickPatch::FrameTime = 0;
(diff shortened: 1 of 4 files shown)


src/Components/Modules/QuickPatch.cpp
```diff
@@ -396,6 +396,24 @@ namespace Components
...
+			for (auto i = 0; i < Game::spawnVars->numSpawnVars; i++)
+			{
+				char** kvPair = Game::spawnVars->spawnVars[i];
+				auto key = kvPair[0];
+				auto val = kvPair[1];
+
+				bool isSpecOps = strncmp(key, "script_specialops", 16) == 0;
+				bool isSpecOpsOnly = *val == 49; // 49 => Ascii "1"
+
+				if (isSpecOps && isSpecOpsOnly) {
```
diamante0018, line 409: Format here is off

src/Components/Modules/QuickPatch.cpp
```diff
@@ -396,6 +396,24 @@ namespace Components
...
+					return true; // This will prevent spawning hopefully;
+				}
+			}
+
+		return Utils::Hook::Call<bool(char*)>(0x444810)(a1);
```
diamante0018, line 414: Can't you do it yourself here instead of calling the game function e.g
`  return !strncmp(a1,"dyn_",4); `
I mean, since you are already hooking the function call, I am simply wondering if it's better to do it like you are doing or call strncmp yourself here. It would not change the code functionality but perhaps it might improve readability since it's not obvious what the function at 0x444810 does (unless you open game binary with ida). Replacing it with `return !strncmp(a1,"dyn_",4);` would make it apparent.

src/Components/Modules/QuickPatch.cpp
```diff
@@ -396,6 +396,24 @@ namespace Components
 		return std::function < T >(reinterpret_cast<T*>(procAddr));
 	}
 
+	bool QuickPatch::IsDynClassnameStub(char* a1) {
```
diamante0018, line 399: format

src/Components/Modules/QuickPatch.cpp
```diff
@@ -396,6 +396,24 @@ namespace Components
...
+			for (auto i = 0; i < Game::spawnVars->numSpawnVars; i++)
+			{
+				char** kvPair = Game::spawnVars->spawnVars[i];
+				auto key = kvPair[0];
+				auto val = kvPair[1];
+
+				bool isSpecOps = strncmp(key, "script_specialops", 16) == 0;
+				bool isSpecOpsOnly = *val == 49; // 49 => Ascii "1"
```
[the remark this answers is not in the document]
Laupetin: Also this only checks if the value *starts with* a 1 character. Practically will work but you could just as well also check for the zero byte for completion sake and to show that it is your intention to check for "1" explicitly.
So when you're at it just do `val[0] == '1' && val[1] == '\0';`

Laupetin (changes requested):
Didn't test the chances for whether they work but if you say they do for your maps then its fine to merge after these few clearity and format things have been addressed. Besides that good pr.
(I know is poorly documented at a lot of places already but that can be addressed in new changes at least)

src/Components/Modules/QuickPatch.cpp
```diff
@@ -396,6 +396,24 @@ namespace Components
...
 	}
 
+	bool QuickPatch::IsDynClassnameStub(char* a1) {
+			for (auto i = 0; i < Game::spawnVars->numSpawnVars; i++)
+			{
+				char** kvPair = Game::spawnVars->spawnVars[i];
+				auto key = kvPair[0];
+				auto val = kvPair[1];
+
+				bool isSpecOps = strncmp(key, "script_specialops", 16) == 0;
```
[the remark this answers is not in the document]
Rackover: i do not know how to count

Rackover:
I updated the PR according to your comments, but I think there's still one thing missing - this code should only execute on 460+ CODO maps, but if I check Zone::Version it's always wrong the first time (but correct when I map restart) due to being set too late (i suppose?). This would probably need addressing otherwise the loading of _any_ map will be slowed down during entity spawning

Laupetin (approved):

Dss0 (approved):

PR #100, opened 2021-07-27
DRAFT: Zonebuilder dump models + materials + techsets?

Rackover:
monkey see, monkey do

most of this is taken from iw3xport, but I tested it and it seems to work - images and models are valid at least.
There is a problem when dumping techset where I hit a memcpy limit, not sure of why, would love to have an opinion!

Also, r_drawmodelnames doesn't suck anymore, and we can dump models out of gfxworld now (very useful to fix iw3 maps and codo maps)


src/Components/Modules/AssetInterfaces/IMaterial.cpp
```diff
@@ -11,6 +11,102 @@ namespace Assets
...
+
+		if (asset->info.name)
+		{
+			buffer.saveString(asset->info.name);
+		}
+
+		if (asset->techniqueSet)
+		{
+			buffer.saveString(asset->techniqueSet->name);
+			//Components::AssetHandler::Dump({ Game::XAssetType::ASSET_TYPE_TECHNIQUE_SET, { asset->techniqueSet } });
```
Laupetin, line 31: Remove commented out code?

src/Components/Modules/AssetInterfaces/IMaterial.cpp
```diff
@@ -11,6 +11,102 @@ namespace Assets
...
+		{
+			for (char i = 0; i < asset->constantCount; ++i)
+			{
+				Game::MaterialConstantDef constantDef;
+				std::memcpy(&constantDef, &asset->constantTable[i], sizeof Game::MaterialConstantDef);
+
+				// This is like the ugliest fix i could come up with.
+				// I'm way to sick to even understand what MaterialConstantDef actually contain
+				// And for now, this shit seems to work
+				if (constantDef.name == "envMapParms"s)
```
Laupetin, line 90: Is this a fix for cod:ol materials? Because this does apply to all materials including vanilla MW2 ones. Also if you would dump and rebuild the zone this would get applied over and over again.
I recommend this gets limited to apply once when it's meant to (so when asset is from cod:ol or whatever)

src/Components/Modules/AssetInterfaces/IMaterialTechniqueSet.cpp
```diff
@@ -10,6 +10,181 @@ namespace Assets
...
+        iw4Decl->name = allocator.duplicateString(Utils::String::VA("iw4xDecl%d", numDecls++));
+        iw4Decl->hasOptionalSource = decl->hasOptionalSource;
+        iw4Decl->streamCount = decl->streamCount;
+
+        memcpy(&iw4Decl->routing, &decl->routing, sizeof(Game::MaterialVertexStreamRouting));
+
+        for (int i = 0; i < iw4Decl->streamCount; i++)
+        {
+            // add in depth destination
+            if (iw4Decl->routing.data[i].dest >= 4)
```
Laupetin, line 160: Again: This applies to all assets not only cod:ol ones

Laupetin (commented):
Limited on my reviewing capabilities because im on phone but got some things I would recommend changing

src/Components/Modules/Maps.cpp
```diff
@@ -237,11 +237,24 @@ namespace Components
 			}
 		}
 
+		if (type == Game::XAssetType::ASSET_TYPE_RAWFILE)
```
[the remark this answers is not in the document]
Rackover: this is the zonebuilder patch!

src/Components/Modules/AssetInterfaces/IMaterial.cpp
Rackover: actually, this code is useful and should not stay commented, but dumping techsets is what makes the game crash in memcpy, that's the part I wanted your opinion on :D (when you'll have a computer).  
Whenever you can test, if you could uncomment this line and run the game with the dump_gfxworld_smodels option so you can have the crash and tell me what you think, that'd be awesome

src/Components/Modules/AssetInterfaces/IMaterial.cpp
Rackover: ah you're correct, they're iw3x fixes i forgot to remove, will look into it. ditto for the next comment

PR #102, opened 2021-07-27
Techset swapping table so we can fix techset issues on codo maps

Rackover:
…on the fly

Just like for iw3->iw4 we'll add more over time if necessary


RektInator:
This should not be merged. This code will cause instability on the client.

RektInator (changes requested):
Shit hook.

PR #104, opened 2021-08-03
Fix codol dynent patch bug

Laupetin:
The patch doesn't respect that the return type should be 32bit BOOL and has a 8bit bool instead.
That causes a game freeze or kick into main menu on loading into private match when in Release mode.
This is due to compiler optimizations not setting the whole 32bit value when returning.


Rackover (approved):

Rackover:
Tested, works

Issue #80, opened 2021-05-28
Game hangs randomly


Rackover:
Alright, I'm gonna try to compile all I have in a single github issue so this become easier to investigate and if anyone has an idea, feel free to post it here.

### The game hangs randomly
IW4x has a glitch that makes the game completely hang - no crash, no stop error. This bug exists since at least the 0.6.1 official release, but might be older. I could not reproduce it on the 0.6.0 release version.

Here are a few key points:
- This bug never happens on the host (whether they're dedicated or listen)
- When this bug happens, it happens for one or more clients, but not necessarily on every client at once
- This bug is very easily reproduceable with BotWarfare, maybe due to the high activity from bots
- This bug happens on vanilla maps (so far Rust, Overgrown, and Terminal, I have not tried reproducing on other maps)
- This bug has a random chance of occuring at any time, but I've had high success reproducing it at round end/start (more details below)
- The game freezes and is unrecoverable without a debugger / memory editor.

Other characteristics that i'm a bit more uncertain about:
- This bug seemingly does not occur on vanilla MW2 nor on other clients afaik (I've failed to reproduce it on aiwnet)
- It has a very low occurence, but once it happened once, it has a higher chance of re-occuring unless the host restarts
- So far, this bug can not be reproduced ever on an AMD CPU.
- This is probably unrelated to Anticheat, as I've managed to reproduced this bug with `DISABLE_ANTICHEAT` defined every single time

### Informations so far
I've reproduced this bug on Terminal and Overgrown and made several dumps using Visual Studio. When the game hangs, the state is usually the following:
- The main thread is stuck inside the `StopSound` subroutine (0x430C55) waiting on mss32.dll to stop all sounds. 
- Another worker thread (which I believe is dedicated to stream sounds from disk and nothing else) is stuck inside the  `SND_ExecuteStreamRead` subroutine (0x64BAFE), waiting on mss32.dll to give a handle to play the sound
- Every other thread is either sleeping, waiting for commands, or in otherwise states that do not sound suspicious to me

The reason these two threads (main thread & the stream worker thread) are waiting on mss32.dll is what makes the game freeze. Looking further, we can see that both of them are stuck in the same part of mss32 - waiting on a mutex for ressource availability. Normally, one thread would lock the mutex first, use the ressource, then free the mutex, and the other thread that would arrive late at the party would then wait for lock, take the ressource once it's available, and free the mutex again.

[...]

PR #106, opened 2021-08-13
Sound mutex workaround

Rackover:
This is a fix for https://github.com/XLabsProject/iw4x-client/issues/80, or rather a workaround.
For reasons explicited in the code, it works "enough" to mitigate the issue and avoid all crashes related to the miles mutex problem, as far as I know.

I've been playing multiple games with this fix for two months with no notable problems, except maybe a slight freeze on round intermission from time to time (was that an avoided crash?). I'm not even sure this was due to the fix or just something that would have occured regardless :D 

So as far as I know this is "safe", but it's also "not great", so - i'm leaving the issue open and hopefully someone can one day come up with a better fix. Kudos to @Laupetin  for helping me investigate this through!

### DIFF
--- a/src/Components/Modules/SoundMutexFix.cpp
+++ b/src/Components/Modules/SoundMutexFix.cpp
@@ -2,42 +2,38 @@
 
 namespace Components
 {
-	/// This component is a workaround for issue https://github.com/XLabsProject/iw4x-client/issues/80
-	/// In case the link goes down, this is a "game hangs randomly" issue:
-	/// 
-	/// Investigations on the issue pointed out it comes from a situation on Intel processors where
-	///		WaitForSingleObjectA is ignored by a thread, for some (?) reason.
-	/// 
-	/// This locks up the game randomly, mostly at the end of rounds or when too many things happen at
-	///		once, due to trying to stop sounds (AIL_Stop_sounds) and playing streams at the same time,
-	///		rushing for the same resource via AIL_lock_mutex. 
-	/// 
-	/// This bug has been reproduced on	mp_terminal, mp_overgrown, mp_rust, with and without bots,
-	///		and so far this has been the only way to circumvent it afaik. This component wraps
-	///		miles' mutex into another mutex, created below, and for some reason (?) that mutex is 
-	///		respected when miles' is not.
-	/// 
-	/// As soon as a real fix is found, please discard this fix. In the meantime, it should not
-	///		have side effects too bad - worst case it might cause a slight performance drop during
-	///		team switch and intermission.
-	/// 
+	// This component is a workaround for issue https://github.com/XLabsProject/iw4x-client/issues/80
+	// In case the link goes down, this is a "game hangs randomly" issue:
+	// 
+	// Investigations on the issue pointed out it comes from a situation on Intel processors where
+	//		WaitForSingleObjectA is ignored by a thread, for some (?) reason.
+	// 
+	// This locks up the game randomly, mostly at the end of rounds or when too many things happen at
+	//		once, due to trying to stop sounds (AIL_Stop_sounds) and playing streams at the same time,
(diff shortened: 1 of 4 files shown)


src/Components/Modules/SoundMutexFix.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
```
Laupetin, line 2: `#include <mutex>` here please :P

src/Components/Modules/SoundMutexFix.hpp
```diff
@@ -0,0 +1,14 @@
...
+namespace Components
+{
+	class SoundMutexFix : public Component
+	{
+	public:
+		SoundMutexFix();
+		~SoundMutexFix();
```
Laupetin, line 9: No need to override the destructor (should have `override` here as well if you do; but you could just delete the destructor. you dont use it anyway)

src/Components/Modules/SoundMutexFix.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+		static void SND_StopStreamChannelHook(int channel);
+		static std::mutex snd_mutex;
```
Laupetin, line 12: You could make the mutex private to prevent hackery

src/Components/Modules/SoundMutexFix.cpp
```diff
@@ -0,0 +1,43 @@
...
+	/// 
+	/// This bug has been reproduced on	mp_terminal, mp_overgrown, mp_rust, with and without bots,
+	///		and so far this has been the only way to circumvent it afaik. This component wraps
+	///		miles' mutex into another mutex, created below, and for some reason (?) that mutex is 
+	///		respected when miles' is not.
+	/// 
+	/// As soon as a real fix is found, please discard this fix. In the meantime, it should not
+	///		have side effects too bad - worst case it might cause a slight performance drop during
+	///		team switch and intermission.
+	///
```
Laupetin, line 23: Thank you for the huge explainatory comment :P Very apprechiated. Just as a little note: MSVC considers three slashes / as a signal for inline documentation. So your huge comment is considered a description for the `snd_mutex` for the IDE (Even though there's a line inbetween).
Results in a huge textbox when hovering over the mutex and is a bit confusing because its not meant as its documentation. (It's nitpicking i know)

src/Components/Modules/SoundMutexFix.cpp
```diff
@@ -0,0 +1,43 @@
...
+	/// 
+
+	std::mutex SoundMutexFix::snd_mutex;
+
+	static void __stdcall LockSoundMutex(int unk)
+	{
+		std::lock_guard lock(SoundMutexFix::snd_mutex);
+
+		DWORD funcPtr = *reinterpret_cast<DWORD*>(0x6D7554); // AIL_close_stream
+		((void(__stdcall*)(int unk))(funcPtr))(unk);
```
Laupetin, line 32: You can also do that via the Utils::Hook api like the other functions so it is better understandable maybe?
Less parenthesis for sure:
```cpp
const DWORD funcPtr = *reinterpret_cast<DWORD*>(0x6D7554); // AIL_close_stream
Utils::Hook::Call<void __stdcall(int unkn)>(funcPtr)(unk);
```

Laupetin (commented):
Like stated before: It's a hacky workaround. It's basically fighting symptoms instead of the cause. We just weren't able to figure out what originally caused it (the freeze is difficult to reproduce making it not easier).
So until a better solution is found make the game at least more playable for people experiencing this freeze.

I added a few comments but its all nothing major so you may decide for or against making this changes. Can be merged as soon as its considered ready by @Rackover

Dss0 (approved):

PR #107, opened 2021-08-14
slowmofix

zweeej:
This is going to fix time lerp in the SetSlowMotion function

on the current develop version the function right away jumps to the entered timescale and ignores the start parameter

### DIFF
--- a/src/Components/Modules/Slowmotion.cpp
+++ b/src/Components/Modules/Slowmotion.cpp
@@ -3,22 +3,35 @@
 namespace Components
 {
 	int SlowMotion::Delay = 0;
-	DWORD applySlowMotionHookLoc = 0x60B2D0;
 
-	void SlowMotion::ApplySlowMotionStub(int timePassed)
+	void SlowMotion::ApplySlowMotion(int timePassed)
 	{
 		if (SlowMotion::Delay <= 0)
 		{
-			__asm
-			{
-				push timePassed
-				call applySlowMotionHookLoc
-				add esp, 4h
-			}
+			Utils::Hook::Call<void(int)>(0x60B2D0)(timePassed);
 		}
 		else
+		{
 			SlowMotion::Delay -= timePassed;
+		}
 	}
+
+	__declspec(naked) void SlowMotion::ApplySlowMotionStub()
+	{
+		__asm
+		{
+			pushad
+			push[esp + 18h]
+
+			call SlowMotion::ApplySlowMotion
+
+			add esp, 4
+			popad
+
+			retn
+		}
+	}
+
 	void SlowMotion::SetSlowMotion()
 	{
 		int duration = 1000;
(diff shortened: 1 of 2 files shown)


Laupetin:
Could you maybe write a few words to what this PR wants to improve?
I fail to see what this commit is supposed to change.

zweeej:
> Could you maybe write a few words to what this PR wants to improve?
> I fail to see what this commit is supposed to change.

This is going to fix the time lerp for the SetSlowMotion gsc function

Laupetin:
Tell me if i'm wrong but for me the functionality looks identical. The way you call the original method is the same as before just with inline assembly.
The stub wrapper is unnecessary i think that is true but it is not hurting either.
So the functionality is identical.

Laupetin:
Also i just tested the slowmotion killcams with the version on develop and they seem to work perfectly fine for me and the person i tested with.
What is true however is that the slowmotion killcams are disabled in gsc in the released iw4x client. That can indeed be addressed since the killcams seem to work fine (It was probably disabled when they were broken and not reenabled).

This would be done in the [rawfile repo](https://github.com/XLabsProject/iw4x-rawfiles). In maps/mp/gametypes/_killcam.gsc the two lines (235 + 237) which set the slowmotion need to be reenabled.
You can do that PR if you want to since you invested time into it :P
Otherwise i would do that myself in a few days if you haven't already.

zweeej:
> Tell me if i'm wrong but for me the functionality looks identical. The way you call the original method is the same as before just with inline assembly.
> The stub wrapper is unnecessary i think that is true but it is not hurting either.
> So the functionality is identical.

It basically has the same functionality no doubt 
but for some reason on the develop version the SetSlowMotion function skips the transition time and directly jumps to the entered timescale

Laupetin:
> > Tell me if i'm wrong but for me the functionality looks identical. The way you call the original method is the same as before just with inline assembly.
> > The stub wrapper is unnecessary i think that is true but it is not hurting either.
> > So the functionality is identical.
> 
> It basically has the same functionality no doubt
> but for some reason on the develop version the SetSlowMotion function skips the transition time and directly jumps to the entered timescale

Okay I see. I didn't understand what you are trying to achieve then sorry. I will check and see if it works like that in a bit then.

Laupetin:
Okay so i tested the patch and it does indeed work. It only works when removing the wrapper **and** using getting rid of the Utils::Hook::Call method in favor of the asm call.

I tried to investigate why this is the case and i gotta say it was a though one but i found out eventually:
The wrapper is necessary because the method apparently relies on certain registers not changing. Therefore the pushad,popad combination saving all register values ensure that even methods that overrides a couple of registers like the Utils::Hook::Call method does not impact the original method.

The wrapper in its original form has an error however in the form of a wrong offset. The correct header should be:
```
__declspec(naked) void SlowMotion::ApplySlowMotionStub()
{
	__asm
	{
		pushad
		push [esp + 18h]

		call SlowMotion::ApplySlowMotion

		add esp,4
		popad

		retn
	}
}
````
Originally it used esp+20h instead of esp+18h which would be correct resulting in using a wrong value as its argument.
The fix can therefore be made by fixing the wrapper alone.

Laupetin:
@zweeej Can you maybe confirm that my suggestion also works for you? If that is the case can you update your PR to fix the wrapper like i mentioned?

The reasoning is that while your fix does in fact fix the problem, it is easy to break it again with minor changes to the ApplySlowMotion method. Also using an inline asm wrapper inside a c method isn't really that desireable.

Kudos for finding the problem, its root and a fix!

zweeej:
> @zweeej Can you maybe confirm that my suggestion also works for you? If that is the case can you update your PR to fix the wrapper like i mentioned?
> 
> The reasoning is that while your fix does in fact fix the problem, it is easy to break it again with minor changes to the ApplySlowMotion method. Also using an inline asm wrapper inside a c method isn't really that desireable.
> 
> Kudos for finding the problem, its root and a fix!

Yes it works perfectly 
thanks for digging deeper into this

Laupetin (approved):
Thank you for your contribution :)

Dss0 (approved):

mjkzy:
> > Tell me if i'm wrong but for me the functionality looks identical. The way you call the original method is the same as before just with inline assembly.
> > The stub wrapper is unnecessary i think that is true but it is not hurting either.
> > So the functionality is identical.
> 
> It basically has the same functionality no doubt
> but for some reason on the develop version the SetSlowMotion function skips the transition time and directly jumps to the entered timescale

when i was working on a mod, we had wonder why it kept directly jumping and after hours of testing, it felt like i achieved just a bit of gradual slow motion by using the function like 50+ times to just to each individual frame

PR #111, opened 2021-08-19
Fix style and KillcamStub

diamante0018:
It worked before because of luck.
Also some commits ago I didn't follow the style so now I fix the style of my previous commits as well.


Laupetin (approved):
Please dont put two patches that do different things independent from another in the same PR anymore in the future :P

diamante0018:
> Please dont put two patches that do different things independent from another in the same PR anymore in the future :P

Will do

Dss0 (approved):

Issue #112, opened 2021-08-19
Server list bugs


Brentdevent:
As seen on the picture people are able to use materials in the server list names + fake player count.

![image](https://user-images.githubusercontent.com/70229620/130120700-ab588f04-0295-4773-9941-4aa146fe7e07.png)

Some people were also complaining about a fatal error (0x0052262B) while in the server list on 18/08/2021.

https://discord.com/channels/219514629703860235/275267435945263104/877672428447363102

![image](https://user-images.githubusercontent.com/70229620/130121166-3b73d6f0-9912-4035-9425-69149eba767b.png)

Perhaps this could be related? I tried to reproduce it myself yesterday but no luck. But they said the issue was resolved today without changing anything.

https://discord.com/channels/219514629703860235/275267435945263104/877910153385619466

Brentdevent:
@.drugS#.0105 was so kind to provide me some crash dumps.

[iw4x-v0.6.0-156-gf3f2c69e-dirty-20210819003107.zip](https://github.com/XLabsProject/iw4x-client/files/7017417/iw4x-v0.6.0-156-gf3f2c69e-dirty-20210819003107.zip)

Brentdevent:
I managed to reproduce the fatal error as well. It seems to occur after a certain amount of servers are loaded and I try to sort on players. 

[iw4x-v0.6.1-20210819221222.dmp.zip](https://github.com/XLabsProject/iw4x-client/files/7017469/iw4x-v0.6.1-20210819221222.dmp.zip)

Dss0:
Addressed in #113 and #114

PR #113, opened 2021-08-20
Fix 2d material text icons

Laupetin:
Make sure that material text icons that are not 2D do not cause issues anymore.

Also strips material text icons entirely for chat messages and server names in the serverlist.

### DIFF
--- a/src/Components/Modules/Dedicated.cpp
+++ b/src/Components/Modules/Dedicated.cpp
@@ -76,6 +76,29 @@ namespace Components
 		}
 	}
 
+	void Dedicated::StripMaterialTextIcons(char* text)
+	{
+		char* currentChar = text;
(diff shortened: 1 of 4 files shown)


diamante0018 (dismissed):
Tested it and it works.

Brentdevent:
Did a quick test. This also seems to fix the fatal error written in issue #112
Good work 👍🏼

diamante0018 (commented):
I tested chat as well, and it works. Quick question: I suppose it's also worth stripping the map name as well?

Laupetin:
> I tested chat as well, and it works. Quick question: I suppose it's also worth stripping the map name as well?

Now everything that is displayed in the server list is stripped from material text icons

Dss0 (approved):

PR #114, opened 2021-08-20
Ignore servers in serverlist that specify an invalid player count

Laupetin:
Filter out any server that answer with a server info that specifies a player count that is greater than 18 or lower than 0.

### DIFF
--- a/src/Components/Modules/ServerList.cpp
+++ b/src/Components/Modules/ServerList.cpp
@@ -472,6 +472,11 @@ namespace Components
 				// Remove server from queue
 				i = ServerList::RefreshContainer.servers.erase(i);
 
+				// Servers with more than 18 players or less than 0 players are faking for sure
+				// So lets ignore those
+				if (server.clients > 18 || server.maxClients > 18 || server.clients < 0 || server.maxClients < 0)
+					return;
+
 				// Check if already inserted and remove
 				auto list = ServerList::GetList();
 				if (!list) return;


diamante0018 (approved):
Tested it, it works.

Rackover (approved):
Very nitpicky: don't we have a macro for max player count?
Shouldn't we have one?

otherwise it's good, and can be merged as soon as you want to :)

src/Components/Modules/ServerList.cpp
```diff
@@ -472,6 +472,11 @@ namespace Components
 				// Remove server from queue
 				i = ServerList::RefreshContainer.servers.erase(i);
 
+				// Servers with more than 18 players or less than 0 players are faking for sure
+				// So lets ignore those
+				if(server.clients > 18 || server.maxClients > 18 || server.clients < 0 || server.maxClients < 0)
```
mjkzy, line 477: add space between ``if`` and the ``(``?

mjkzy (changes requested):

src/Components/Modules/ServerList.cpp
Laupetin: and i thought i was nitpicky lmao

Laupetin:
> Very nitpicky: don't we have a macro for max player count?
> Shouldn't we have one?
> 
> otherwise it's good, and can be merged as soon as you want to :)

Not macro. The game for sure has one the client doesn't though.

Dss0 (approved):

PR #123, opened 2021-08-29
[Dvar] Use set raw method on dvars by default

diamante0018:
Setraw functions have been removed of course

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -88,7 +88,7 @@ namespace Components
 	{
 		if (this->dvar && this->dvar->name)
 		{
-			Game::Dvar_SetCommand(this->dvar->name, string);
+			this->dvar->current.string = string;
 		}
 	}
 	void Dvar::Var::set(const std::string& string)
@@ -99,30 +99,21 @@ namespace Components
 	{
 		if (this->dvar && this->dvar->name)
 		{
-			Game::Dvar_SetCommand(this->dvar->name, Utils::String::VA("%i", integer));
+			this->dvar->current.integer = integer;
 		}
 	}
 	void Dvar::Var::set(float value)
 	{
 		if (this->dvar && this->dvar->name)
 		{
-			Game::Dvar_SetCommand(this->dvar->name, Utils::String::VA("%f", value));
-		}
-	}
-
-	void Dvar::Var::setRaw(int integer)
-	{
-		if (this->dvar)
-		{
-			this->dvar->current.integer = integer;
-		}
-	}
-
-	void Dvar::Var::setRaw(float value)
-	{
-		if (this->dvar)
-		{
 			this->dvar->current.value = value;
+		}
+	}
+	void Dvar::Var::set(bool enabled)
+	{
+		if (this->dvar && this->dvar->name)
+		{
+			this->dvar->current.enabled = enabled;
 		}
 	}
 
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -20,6 +20,7 @@ namespace Components
 	{
 		return this->dvar;
 	}
+
 	template <> char* Dvar::Var::get()
 	{
 		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_STRING && this->dvar->current.string)
@@ -29,10 +30,12 @@ namespace Components
 
 		return const_cast<char*>("");
 	}
+
 	template <> const char* Dvar::Var::get()
 	{
 		return this->get<char*>();
 	}
+
 	template <> int Dvar::Var::get()
 	{
 		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_INT)
@@ -42,10 +45,12 @@ namespace Components
 
 		return 0;
 	}
+
 	template <> unsigned int Dvar::Var::get()
 	{
 		return static_cast<unsigned int>(this->get<int>());
 	}
+
 	template <> float Dvar::Var::get()
 	{
 		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_FLOAT)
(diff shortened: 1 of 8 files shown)


src/Game/Functions.cpp
```diff
@@ -110,6 +110,9 @@ namespace Game
 	Dvar_InfoString_Big_t Dvar_InfoString_Big = Dvar_InfoString_Big_t(0x4D98A0);
 	Dvar_SetCommand_t Dvar_SetCommand = Dvar_SetCommand_t(0x4EE430);
 
+	Free_String_t Free_String = Free_String_t(0x470E80);
+	Copy_String_t Copy_String = Copy_String_t(0x4F3C80);
```
Laupetin, line 114: I think those are not needed since you dont use them

src/Game/Structs.hpp
```diff
@@ -90,6 +90,7 @@ namespace Game
 		DVAR_FLAG_SERVERINFO = 0x400,		//in the getstatus oob
 		DVAR_FLAG_WRITEPROTECTED = 0x800,		//write protected
 		DVAR_FLAG_UNKNOWN1000 = 0x1000,		//unknown
+		DVAR_FLAG_UNKNOWN1800 = 0x1800,		//unknown
```
Laupetin, line 93: This is first of all not used and also not a flag but a combination of the flag values 0x1000 and 0x800

src/Game/Functions.hpp
```diff
@@ -256,6 +259,12 @@ namespace Game
 	typedef dvar_t* (__cdecl * Dvar_SetCommand_t)(const char* name, const char* value);
 	extern Dvar_SetCommand_t Dvar_SetCommand;
 
+	typedef void (__cdecl* Free_String_t)(const char* string);
+	extern Free_String_t Free_String;
+
+	typedef const char* (__cdecl* Copy_String_t)(const char* string);
+	extern Copy_String_t Copy_String;
+
```
Laupetin, line 267: Not used

src/Game/Structs.hpp
```diff
@@ -119,7 +120,7 @@ namespace Game
 		DVAR_SOURCE_DEVGUI = 0x3,
 	};
 
-	typedef enum
+	typedef enum : char
```
Laupetin, line 123: Didn't you want to change the dvar struct to use this enum then? Because what's the point of doing this is you are not making dvar_t::type this enum type

src/Components/Modules/Dvar.cpp
```diff
@@ -97,32 +97,26 @@ namespace Components
...
+			this->dvar->current.value = value;
+			this->dvar->latched.value = value;
 		}
 	}
-
-	void Dvar::Var::setRaw(float value)
+	void Dvar::Var::set(bool enabled)
 	{
-		if (this->dvar)
+		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_BOOL)
```
Laupetin, line 116: i would add an assert here (before this line) that checks for the dvar type to not silently fail in debug mode. 
```cpp
assert(this->dvar->type == Game::DVAR_TYPE_BOOL);
```

src/Components/Modules/Dvar.cpp
```diff
@@ -97,32 +97,26 @@ namespace Components
...
-		if (this->dvar && this->dvar->name)
-		{
-			Game::Dvar_SetCommand(this->dvar->name, Utils::String::VA("%f", value));
-		}
-	}
-
-	void Dvar::Var::setRaw(int integer)
-	{
-		if (this->dvar)
+		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_FLOAT)
```
Laupetin, line 108: I would add an assert here (see ``void Dvar::Var::set(bool enabled)``)

src/Components/Modules/Dvar.cpp
```diff
@@ -97,32 +97,26 @@ namespace Components
 	}
 	void Dvar::Var::set(int integer)
 	{
-		if (this->dvar && this->dvar->name)
+		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_INT)
```
Laupetin, line 100: I would add an assert here (see ``void Dvar::Var::set(bool enabled)``)

src/Components/Modules/Dvar.cpp
```diff
@@ -86,9 +86,9 @@ namespace Components
 	}
 	void Dvar::Var::set(const char* string)
 	{
-		if (this->dvar && this->dvar->name)
+		if (this->dvar && this->dvar->type == Game::dvar_type::DVAR_TYPE_STRING)
```
Laupetin, line 89: I would add an assert here (see ``void Dvar::Var::set(bool enabled)``)

Laupetin (changes requested):

src/Components/Modules/Dvar.cpp
```diff
@@ -97,32 +98,26 @@ namespace Components
...
+		assert(this->dvar->type == Game::DVAR_TYPE_FLOAT);
 		if (this->dvar)
 		{
-			this->dvar->current.integer = integer;
+			Game::Dvar_SetFloat(this->dvar, value);
 		}
 	}
-
-	void Dvar::Var::setRaw(float value)
+	void Dvar::Var::set(bool enabled)
```
Laupetin, line 115: i mean since you keep bugging people because of code style: empty line between methods would be apprechiate. in reality i dont care though

Laupetin (approved):
looks fine. there might be a few high performant situations where you wouldn't necessarly want to use the game's Dvar_SetX functions because they convert the value into a string and then parse the string again but in the places that are effected by this PR there is no such case.

src/Components/Modules/Dvar.cpp
diamante0018: It wasn't me who bothered you for the space between if and ( 😂

Dss0 (approved):

PR #124, opened 2021-09-05
[Structs] Update client_t

diamante0018:


src/Game/Structs.hpp
```diff
@@ -4215,6 +4215,51 @@ namespace Game
...
+	{
+		char __pad0[0x5E0];
+	};
+
+	static_assert(sizeof(netProfileInfo_t) == 0x5E0);
+
+	struct netchan_t
+	{
+		// 0
+		int outgoingSequence;
```
Laupetin, line 4228: i would put the offset comments on the side of the field. otherwise they take a lot of space for no reason and i like to quickly scroll over code :P

Laupetin: Also this struct seems to be pretty much fully reversed. are the offsets still necessary? most of the time they are just a help if the struct is still being worked on.

src/Game/Structs.hpp
```diff
@@ -4624,47 +4669,54 @@ namespace Game
...
 		// 278272
-		unsigned __int64 steamid;
+		unsigned __int64 steamID;
 		// 278280
-		char pad7[403592];
+		char __pad10[403592];
 	} client_t;
 #pragma pack(pop)
 
+	static_assert(sizeof(client_t) == 681872);
```
Laupetin, line 4718: idk if the static asserts are a mistake but by me you can leave them in. Would make sense to confirm the struct size if you know it for sure.
However please use a hex number here for consistency

Laupetin (commented):
Struct looks fine by me. Just minor style changes. I confirmed a couple of offsets and sizes but didn't test it by running the game a bit with it. i'll leave that to you

src/Game/Structs.hpp
diamante0018: Okay, I will remove them.

Dss0 (approved):

PR #125, opened 2021-09-05
Allow loading assets from disk even if they're already in memory

Rackover:


Rackover:
Some models bear the same name in iw4 than in iw3, but aren't the same! These causes issues in iw3-ported maps because zonebuilder takes memory assets over disk assets. Here's an example:

IW3
image


IW4x
image


IW4x (fixed)
image

Laupetin (approved):
Can't say much about the cod4 material mapping aspects since i don't work with that. Code looks okay though.

PR #129, opened 2021-09-12
[Network] Check for invalid reliableAcknowledge

diamante0018:


src/Components/Modules/Network.cpp
```diff
@@ -342,6 +342,18 @@ namespace Components
 			Game::NET_DeferPacketToClient(from, msg);
 	}
 
+	void Network::SV_ExecuteClientMessageStub(Game::client_t* client, Game::msg_t* msg)
+	{
+		if (client->reliableAcknowledge < 0)
+		{
+			client->reliableAcknowledge = client->reliableSequence;
+			Network::SendCommand(Game::NS_SERVER, client->netchan.remoteAddress, "error", "EXE_LOSTRELIABLECOMMANDS");
```
Laupetin, line 350: Maybe you could add a log message here to log on serverside that a player has been kicked due to negative reliableAcknowledge

src/Game/Structs.hpp
```diff
@@ -4220,8 +4220,6 @@ namespace Game
 		char __pad0[0x5E0];
 	};
 
-	static_assert(sizeof(netProfileInfo_t) == 0x5E0);
```
Laupetin, line 4223: Why remove this static assert now. Now either leave them in or not but don't remove and add those please. I'd prefer if you readded this if the size was confirmed to be true

src/Game/Structs.hpp
```diff
@@ -4662,15 +4660,27 @@ namespace Game
...
+		int lastConnectTime; // 135868
 		int snapNum; // 135872
-		int __pad6; // 135876
+		int __pad5; // 135876
 		short ping; // 135880
-		char __pad7[133158]; // 135882
+		char __pad6[14]; // 135882
+		int pureAuthentic; // 135896
+		char __pad7[133138]; // 135900
+		short scriptID; // 269038
```
Laupetin, line 4680: Where are the fields used. What's the point of doing this?

Laupetin (commented):
Looks okay to me. Tested it and does the job it's supposed to. Add a log message upon kick though please so it's possible to see that this logic was executed in the server log. Helps in case this does not work as intended.

src/Components/Modules/Network.cpp
diamante0018: Will do

src/Game/Structs.hpp
diamante0018: I thought it was a redundant static assert since the only field in the struct is a char array

src/Game/Structs.hpp
diamante0018: I have the time to reverse it, that's all

Laupetin (approved):

Rackover (approved):

PR #130, opened 2021-09-19
Fonticons for 2D text rendering

Laupetin:
While it was possible before to make use of materials in e.g. chat messages or custom titles the usage of them was not practical, pretty hard and required a deep knowledge of the text rendering algorithm (^\x01 + width + height + materialnamelen + materialname) this PR tries to introduce a userfriendly and usable way of achieving the same effect which is using materials in text.

# Features

## Rendering of font icons
This PR introduces "font icons" which is a way to use certain predefined materials in texts like chat messages with the simple to use syntax ``:fonticonname:`` (inspired by the syntax of emojis on certain chat platforms like discord).

fonticons1

The usage only works with a preselected list of materials to be able to exclude materials that would not make sense to be used as a font icon (Like a "messagebox dropshadow texture" or something that is unclear what it is supposed to represent) or would be annoying for other players (like a meters long compass texture).

The specification of usable font icons is done with ``mp/fonticons.csv`` and includes a name for a fonticon and the respective material name.

## Font icon autocomplete

To avoid having to know all fonticon names by heart there is an autocompletion dialog that displays when you try to write a font icon.

fonticons2

This autocompletion dialog is implemented for chat and console. It is only displayed when a double colon and an alphabetic character are typed with a space or symbol before.

In case a user does not like the dialog, it can be turned off with ``cg_fontIconAutocomplete 0``. A smaller version of the dialog without the hint text can be accessed by using ``cg_fontIconAutocompleteHint 0``.

## Font icon modifiers

Additionally to the name of the fonticons, modifiers can be specified. Modifiers are specified using a single respective character after a ``+`` separator behind the fonticon name. (``:fonticonname+b:`` would use the ``b`` modifier).

There are currently 3 modifiers implemented:
- h: Mirrors the icon horizontally
- v: Mirrors the icon vertically
- b: Makes the icon bigger (by 50%)

fonticons3

There can be multiple modifiers specified.

## Visible Fonticon and Color escape sequence editing

When going into a color escape sequence or a fonticon with the text cursor, they are rendered as text, making editing them more comfortable.

fonticons4
fonticons5

## Hudicon and fonticon stripping

This PR also brings internal changes to stripping of hudicons and fonticons so they are stripped entirely instead of just stripping the first two characters to break the escape sequence.

# Testing

This patch requires the fonticon csv stringtable to be available. A draft version of it can be found [here as a PR to the rawfile repository](https://github.com/XLabsProject/iw4x-rawfiles/pull/7). 
I plan on changing it a bit still to change the names of some fonticons to versions that might make more sense for players. For testing the version on the draft can be used without problems though.

To be able to test this, put the ``iw4x/iw4x_00/mp/fonticons.csv`` file in ``userraw/mp/fonticons.csv`` of your game folder (for example).


src/Components/Modules/Chat.cpp
```diff
@@ -0,0 +1,205 @@
...
+			mov al, Chat::SendChat
+
+			test al, al
+			jnz return
+
+			// Don't send the chat
+			pop eax
+			retn
+
+			return:
```
diamante0018, line 72: Visual studio is the culprit but the return label was originally one tab to the left.

src/Components/Modules/Chat.cpp
```diff
@@ -0,0 +1,205 @@
...
+		const auto chatWidth = static_cast<float>(cg_chatWidth.get<int>());
+		const auto chatTime = (*cg_chatTime)->current.integer;
+		if (chatHeight < 0 || static_cast<unsigned>(chatHeight) > std::extent_v<decltype(Game::cgs_t::teamChatMsgs)> || chatWidth <= 0 || chatTime <= 0)
+		{
+			Game::cgsArray[0].teamLastChatPos = 0;
+			Game::cgsArray[0].teamChatPos = 0;
+			return;
+		}
+
+        TextRenderer::FontIconInfo fontIconInfo{};
```
diamante0018, line 124: One missing tab here?

src/Components/Modules/Chat.cpp
```diff
@@ -0,0 +1,205 @@
...
+				if (text[0] == ' ')
+					lastSpace = p;
+				*p++ = *text++;
+				len += 1.0f;
+			}
+		}
+		*p = 0;
+		Game::cgsArray[0].teamChatMsgTimes[Game::cgsArray[0].teamChatPos % chatHeight] = Game::cgArray[0].time;
+		if (Game::cgsArray[0].teamChatPos++ + 1 - Game::cgsArray[0].teamLastChatPos > chatHeight)
+            Game::cgsArray[0].teamLastChatPos = Game::cgsArray[0].teamChatPos + 1 - chatHeight;
```
diamante0018, line 175: Space vs Tab

src/Components/Modules/PlayerName.cpp
```diff
@@ -2,37 +2,77 @@
...
+            char nameBuffer[64] = { 0 };
+            TextRenderer::StripColors(name, nameBuffer, sizeof(nameBuffer));
+            TextRenderer::StripAllTextIcons(nameBuffer, buffer, size);
+        }
+        else
+        {
+            TextRenderer::StripAllTextIcons(name, buffer, size);
+        }
+
+        std::string readablePlayerName(buffer);
```
diamante0018, line 20: I am aware it is possible to pass to the string constructor the size of the source buffer which I've seen momo do very often, perhaps as a precaution? Would it fit well here?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            ++in;
+        }
+        *out = '\0';
+    }
+
+    std::string TextRenderer::StripColors(const std::string& in)
+    {
+        char buffer[1000] = { 0 }; // Should be more than enough
+        StripColors(in.data(), buffer, sizeof(buffer));
+        return std::string(buffer);
```
diamante0018, line 1152: As above, pass the size of the source buffer to the string constructor?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+
+        }
+        *out = '\0';
+    }
+
+    std::string TextRenderer::StripMaterialTextIcons(const std::string& in)
+    {
+        char buffer[1000] = { 0 }; // Should be more than enough
+        StripAllTextIcons(in.data(), buffer, sizeof(buffer));
+        return std::string(buffer);
```
diamante0018, line 1199: See above

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            ++in;
+        }
+        *out = '\0';
+    }
+
+    std::string TextRenderer::StripAllTextIcons(const std::string& in)
+    {
+        char buffer[1000] = { 0 }; // Should be more than enough
+        StripAllTextIcons(in.data(), buffer, sizeof(buffer));
+        return std::string(buffer);
```
diamante0018, line 1256: See above

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        Utils::Hook(0x5A50BB, Con_DrawInput_Hk, HOOK_CALL).install()->quick();
+
+        // Handle key inputs for console and chat
+        Utils::Hook(0x4F685C, Console_Key_Hk, HOOK_CALL).install()->quick();
+        Utils::Hook(0x4F6694, Message_Key_Stub, HOOK_CALL).install()->quick();
+        Utils::Hook(0x4F684C, Message_Key_Stub, HOOK_CALL).install()->quick();
+
+        PatchColorLimit(COLOR_LAST_CHAR);
+    }
+}
```
diamante0018, line 1493: New line

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
+#pragma once
+
+namespace Components
+{
+	enum TextColor
+	{
+	    TEXT_COLOR_BLACK = 0,
```
diamante0018, line 7: Space vs Tab

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
...
+		TEXT_COLOR_BLUE = 4,
+		TEXT_COLOR_LIGHT_BLUE = 5,
+		TEXT_COLOR_PINK = 6,
+		TEXT_COLOR_DEFAULT = 7,
+		TEXT_COLOR_AXIS = 8,
+		TEXT_COLOR_ALLIES = 9,
+		TEXT_COLOR_RAINBOW = 10,
+		TEXT_COLOR_SERVER = 11, // using that color in infostrings (e.g. your name) fails, ';' is an illegal character!
+
+		TEXT_COLOR_COUNT
```
diamante0018, line 20: Why new line above this?

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
...
+			unsigned char s;
+			unsigned char v;
+		};
+
+		enum FontIconAutocompleteInstance
+		{
+		    FONT_ICON_ACI_CONSOLE,
+		    FONT_ICON_ACI_CHAT,
+
+			FONT_ICON_ACI_COUNT
```
diamante0018, line 64: Space vs Tab + Random newline?

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
...
+		static void PatchColorLimit(char limit);
+		static bool Dvar_GetUnpackedColorByName(const char* name, float* expandedColor);
+		static void GetUnpackedColorByNameStub();
+
+		static Game::GfxImage* GetFontIconColorMap(const Game::Material* fontIconMaterial);
+		static float DrawFontIcon(const FontIconInfo& fontIcon, float x, float y, float sinAngle, float cosAngle, const Game::Font_s* font, float xScale, float yScale, unsigned color);
+
+		static float GetMonospaceWidth(Game::Font_s* font, int rendererFlags);
+		static void GlowColor(Game::GfxColor* result, Game::GfxColor baseColor, Game::GfxColor forcedGlowColor, int renderFlags);
+        static unsigned R_FontGetRandomLetter(int seed);
```
diamante0018, line 205: Tab here

src/Game/Functions.cpp
```diff
@@ -1315,22 +1346,123 @@ namespace Game
...
 			pop eax
-			retn
+			ret
+		}
+	}
+
+	__declspec(naked) void RB_DrawChar(Material* /*material*/, float /*x*/, float /*y*/, float /*w*/, float /*h*/, float /*sinAngle*/, float /*cosAngle*/, Glyph* /*glyph*/, unsigned int /*color*/)
+	{
+	    __asm
+		{
```
diamante0018, line 1401: This curly bracket is meant to be one tab to the left, isn't it?

src/Game/Functions.cpp
```diff
@@ -1315,22 +1346,123 @@ namespace Game
...
+
+			popad
+			ret
+		}
+	}
+
+	__declspec(naked) void RB_DrawStretchPicRotate(Material* /*material*/, float /*x*/, float /*y*/, float /*w*/, float /*h*/, float /*s0*/, float /*t0*/, float /*s1*/, float /*t1*/, float /*sinAngle*/, float /*cosAngle*/, unsigned int /*color*/)
+	{
+	    __asm
+		{
```
diamante0018, line 1426: See above

src/Components/Modules/Chat.cpp
```diff
@@ -0,0 +1,205 @@
...
+				p[0] = '^';
+				p[1] = text[1];
+				lastColor = ColorIndexForChar(text[1]);
+				p += 2;
+				text += 2;
+			}
+			else
+			{
+				if (text[0] == ' ')
+					lastSpace = p;
```
diamante0018, line 167: Tab

src/Components/Modules/PlayerName.cpp
```diff
@@ -2,37 +2,77 @@
...
-	std::string PlayerName::PlayerNames[18];
+    Dvar::Var PlayerName::sv_allowColoredNames;
 
-	int PlayerName::GetClientName(int /*localClientNum*/, int index, char *buf, int size)
+    void PlayerName::UserInfoCopy(char* buffer, const char* name, const size_t size)
+    {
+        if (!sv_allowColoredNames.get<bool>())
+        {
+            char nameBuffer[64] = { 0 };
+            TextRenderer::StripColors(name, nameBuffer, sizeof(nameBuffer));
```
diamante0018, line 12: Perhaps I didn't understand how the underlying function works but why was sizeof used here and not the size parameter?

src/Components/Modules/PlayerName.cpp
```diff
@@ -2,37 +2,77 @@
...
+        {
+            TextRenderer::StripAllTextIcons(name, buffer, size);
+        }
+
+        std::string readablePlayerName(buffer);
+        readablePlayerName = Utils::String::Trim(readablePlayerName);
+
+        if (readablePlayerName.size() < 3)
+        {
+            strncpy(buffer, "Unknown Soldier", size);
```
diamante0018, line 25: strncpy_s here ?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            call ChatHandleKeyDown
+            add esp, 0x8
+            test al,al
+            jnz skipHandling
+
+            popad
+            call Message_Key
+            ret
+
+        skipHandling:
```
diamante0018, line 514: Label should be a bit more to the left ?

src/Components/Modules/TextRenderer.hpp
diamante0018, line 7: Tab here

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
...
+		TEXT_COLOR_YELLOW = 3,
+		TEXT_COLOR_BLUE = 4,
+		TEXT_COLOR_LIGHT_BLUE = 5,
+		TEXT_COLOR_PINK = 6,
+		TEXT_COLOR_DEFAULT = 7,
+		TEXT_COLOR_AXIS = 8,
+		TEXT_COLOR_ALLIES = 9,
+		TEXT_COLOR_RAINBOW = 10,
+		TEXT_COLOR_SERVER = 11, // using that color in infostrings (e.g. your name) fails, ';' is an illegal character!
+
```
diamante0018, line 19: Reason for a newline here?

src/Game/Functions.cpp
```diff
@@ -1315,22 +1346,123 @@ namespace Game
 		}
 	}
 
-	__declspec(naked) Glyph* R_GetCharacterGlyph(Font_s* /*font */, unsigned int /*letter*/)
+	__declspec(naked) Glyph* R_GetCharacterGlyph(Font_s* /*font*/, unsigned int /*letter*/)
 	{
 	    __asm
 		{
```
diamante0018, line 1352: Shouldn't these curly brackets (and for the following functions) be one tab to the left?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        if(entry.material == nullptr)
+        {
+            auto* materialEntry = Game::DB_FindXAssetEntry(Game::XAssetType::ASSET_TYPE_MATERIAL, entry.materialName.data());
+            if (materialEntry == nullptr)
+                return false;
+            auto* material = materialEntry->asset.header.material;
+            if (material == nullptr || material->techniqueSet == nullptr || material->techniqueSet->name == nullptr)
+                return false;
+
+            if(strcmp(material->techniqueSet->name, "2d") != 0)
```
diamante0018, line 625: Use utils string starts_with here?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        {
+            if (text[i] == 0)
+                return 0;
+        }
+
+        const std::string materialName(text, materialNameLen);
+        text += materialNameLen;
+
+        auto* material = Game::DB_FindXAssetHeader(Game::XAssetType::ASSET_TYPE_MATERIAL, materialName.data()).material;
+        if (material == nullptr || material->techniqueSet == nullptr || material->techniqueSet->name == nullptr || strcmp(material->techniqueSet->name, "2d") != 0)
```
diamante0018, line 764: utils string start with here (for "2d")?

Laupetin: Why that? I wanna make sure it's using the 2d techset and not a techset that has a name that starts with 2d? There's probably none even.

src/Components/Modules/TextRenderer.cpp
Laupetin: Why that? I wanna make sure it's using the 2d techset and not a techset that has a name that starts with 2d? There's probably none even.

src/Game/Functions.cpp
Laupetin: yes my VS does that for auto code formatting for some reason. Another reason to use automatic formatting :P

src/Components/Modules/TextRenderer.hpp
Laupetin: because the count entry is not part of the actual enum but a way to access the number of enum members / the index after the last enum entry

src/Components/Modules/PlayerName.cpp
Laupetin: You only need to specify a length if you do not want the std::string to be constructed from the entire string but a substring. If no length is specified it uses the entire string until a zero byte is encountered. Not a precaution but a different usecase.

src/Components/Modules/TextRenderer.cpp
Laupetin: see above

src/Components/Modules/TextRenderer.cpp
Laupetin: ditto

src/Components/Modules/TextRenderer.hpp
Laupetin: again because this is not part of the enum members but a way to access the index after the last entry / the count of enum members. This is also why this has no specified value

src/Components/Modules/TextRenderer.hpp
Laupetin: not random

src/Components/Modules/PlayerName.cpp
Laupetin: Because this is reading from name and writing to the temporary name buffer. The size parameter is there to make sure the output buffer does not overflow therefore the size of the temporary buffer is needed here

src/Components/Modules/TextRenderer.hpp
Laupetin: yes

Rackover (commented):
It's a gigantic amount of code and a very nice feature, very well executed, I find little to criticize :)
Most of the comments are probably a bit naive, but here are two feedbacks I consider important:

- The text in your small menu should be localizable, so added to your rawfile PR and used as such
- Magic chars like `:` and `+` and `b` that you use for declaring or modifying emojis deserve to be constants, so we can change them in the future if we ever want :) (plus its good practice)

src/Components/Modules/PlayerName.cpp
Laupetin: Doesn't make a difference here since the size of the buffer is ``size``. So if strncpy_s was used then buffer size and copy count would be identical. It's also not possible to use the implicit buffer size because its a pointer. So not necessary or helpful here imo

src/Components/Modules/Chat.cpp
```diff
@@ -0,0 +1,205 @@
...
+				*p++ = *text++;
+				len += 1.0f;
+			}
+		}
+		*p = 0;
+		Game::cgsArray[0].teamChatMsgTimes[Game::cgsArray[0].teamChatPos % chatHeight] = Game::cgArray[0].time;
+		if (Game::cgsArray[0].teamChatPos++ + 1 - Game::cgsArray[0].teamLastChatPos > chatHeight)
```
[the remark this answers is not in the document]
Laupetin: yeah you are right. was too focused on recreating what the game does that i forgot to make it readable

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        h = hsv.h;
+        s = hsv.s;
+        v = hsv.v;
+
+        region = static_cast<uint8_t>(h / 43);
+        remainder = (h - (region * 43)) * 6;
+
+        p = static_cast<uint8_t>((v * (255 - s)) >> 8);
+        q = static_cast<uint8_t>((v * (255 - ((s * remainder) >> 8))) >> 8);
+        t = static_cast<uint8_t>((v * (255 - ((s * (255 - remainder)) >> 8))) >> 8);
```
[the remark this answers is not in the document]
Laupetin: I didn't actually write this code. I copied it over from the Colors patch since this was better located in the text renderer patch.
Gotta be honest i'm also not sure why these characters were used the mathmatical symbols here. The calculation itself is probably based on https://en.wikipedia.org/wiki/HSL_and_HSV#Color_conversion_formulae or similar.
I would leave it like this tbh because i don't think this can really be refactored into something that is understandable without looking at the maths behind it.
Is that okay?

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+                    context.results[context.resultCount++] = {
+                        Utils::String::VA(":%s:", fontIconEntry.iconName.data()),
+                        fontIconEntry.iconName
+                    };
+                }
+                else
+                    context.hasMoreResults = true;
+            }
+            else if (compareValue > 0)
+                break;
```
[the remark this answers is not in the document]
Laupetin: idk i did that on purpose. I'm usually okay with leaving away the curly braces if either all branches dont use them or if its a one liner that has no branches after it that use curly braces.
But that's honestly up to code style and idk if one way to do it is better than the other way.
I would leave it like this and maybe look into having automated code formatting in the near future (I also dont feel like having tabs vs spaces crap in the code all the time as well so its kind of necessary to do anyway)

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+
+    void TextRenderer::UpdateAutocompleteContext(TextRenderer::FontIconAutocompleteContext& context, const Game::field_t* edit, Game::Font_s* font, const float textXScale)
+    {
+        int fontIconStart = -1;
+        auto inModifiers = false;
+
+        for(auto i = 0; i < edit->cursor; i++)
+        {
+            const auto c = static_cast<unsigned char>(edit->buffer[i]);
+            if (c == ':')
```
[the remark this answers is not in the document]
Laupetin: I guess yea. Will do

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+                {
+                    fontIconStart = -1;
+                    inModifiers = false;
+                }
+            }
+        }
+
+        if(fontIconStart < 0 // Not in fonticon sequence
+            || fontIconStart == edit->cursor // Did not type the first letter yet
+            || !isalpha(static_cast<unsigned char>(edit->buffer[fontIconStart])) // First letter of the icon is not alphabetic
```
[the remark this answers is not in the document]
Laupetin: Thing is we do not use iwi names as fonticon names. Also not material names. The fonticon names are part of the fonticons.csv and therefore can be done in a way that no fonticon starts with a number or symbol (which i did for the current draft).
I think there was only one material that started with a letter (8ball) and i probably made it eightball or something like that.
I did this because i noticed that there are cases in which you do not want a fonticon popup but you do. In this case it was hostname:port for the connect command.

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        }
+
+        // Update results for query and scroll
+        context.lastQuery = std::string(&edit->buffer[fontIconStart], edit->cursor - fontIconStart);
+        UpdateAutocompleteContextResults(context, font, textXScale);
+    }
+
+    void TextRenderer::DrawAutocompleteModifiers(const FontIconAutocompleteContext& context, const float x, const float y, Game::Font_s* font, float textXScale, float textYScale)
+    {
+        const auto* text = "The following modifiers are available:\n"
```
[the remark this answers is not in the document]
Laupetin: Agreed (It will not use printf args but localized strings have their own way of inserting variables into it. ``&&1`` i think? i forgot).

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            else
+                Game::R_AddCmdDrawText(result.materialName.c_str(), INT_MAX, font, x + context.maxFontIconWidth + colSpacing, currentY, textXScale, textYScale, 0.0, TEXT_COLOR, 0);
+            currentY += lineHeight;
+        }
+
+        if(hintEnabled)
+        {
+            Game::R_AddCmdDrawText("Press ^3TAB ^7for autocomplete", INT_MAX, font, x, currentY, textXScale, textYScale, 0.0, HINT_COLOR, 0);
+            currentY += lineHeight;
+            Game::R_AddCmdDrawText("Use ^3+ ^7for modifiers", INT_MAX, font, x, currentY, textXScale, textYScale, 0.0, HINT_COLOR, 0);
```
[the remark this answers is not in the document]
Laupetin: oke

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        currentY += lineHeight;
+
+        const auto selectedIndex = context.selectedOffset - context.resultOffset;
+        for(auto resultIndex = 0u; resultIndex < context.resultCount; resultIndex++)
+        {
+            const auto& result = context.results[resultIndex];
+            Game::R_AddCmdDrawText(result.fontIconName.c_str(), INT_MAX, font, x, currentY, textXScale, textYScale, 0.0, TEXT_COLOR, 0);
+
+            if(selectedIndex == resultIndex)
+                Game::R_AddCmdDrawText(Utils::String::VA("^2%s", result.materialName.c_str()), INT_MAX, font, x + context.maxFontIconWidth + colSpacing, currentY, textXScale, textYScale, 0.0, TEXT_COLOR, 0);
```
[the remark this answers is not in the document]
Laupetin: you got me

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        auto yy = static_cast<float>(y);
+        yy += static_cast<float>(Game::R_TextHeight(font)) * normalizedScale * 1.5f;
+        auto ww = normalizedScale;
+        auto hh = normalizedScale;
+        Game::ScrPlace_ApplyRect(screenPlacement, &xx, &yy, &ww, &hh, horzAlign, vertAlign);
+
+        UpdateAutocompleteContext(autocompleteContext, edit, font, ww);
+        if (autocompleteContext.autocompleteActive)
+        {
+            DrawAutocomplete(autocompleteContext, std::floor(xx), /*std::floor(*/ yy/*)*/, font, ww, hh);
```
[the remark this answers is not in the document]
Laupetin: whoops

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            AutocompleteUp(context);
+            return true;
+
+        case Game::K_DOWNARROW:
+        case Game::K_KP_DOWNARROW:
+            AutocompleteDown(context);
+            return true;
+
+        case Game::K_TAB:
+            AutocompleteFill(context, scrPlace, edit);
```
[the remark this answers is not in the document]
Laupetin: Idk if ENTER is really good here because if the autocomplete pops up you might not intend to use its correction and send the message anyway (for example if you are not typing a font icon but just happen to use a ``:`` before something. I guess its okay for chat because you can still do ESC. But that doesn't work for console because the key is not delivered to the hooked location so i dont really want to do it there.

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        
+        if(*curPos == '+')
+        {
+            auto breakArgs = false;
+            while(!breakArgs)
+            {
+                curPos++;
+                switch(*curPos)
+                {
+                case 'h':
```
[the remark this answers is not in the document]
Laupetin: yes

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+        return static_cast<float>(font->pixelHeight) * (colWidth / colHeight) * xScale * sizeMultiplier;
+    }
+
+    float TextRenderer::DrawFontIcon(const FontIconInfo& fontIcon, const float x, const float y, const float sinAngle, const float cosAngle, const Game::Font_s* font, const float xScale, const float yScale, const unsigned color)
+    {
+        const auto* colorMap = GetFontIconColorMap(fontIcon.material);
+        if (colorMap == nullptr)
+            return 0;
+
+        float s0, t0, s1, t1;
```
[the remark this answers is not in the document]
Laupetin: That's the naming convention of the game. idk.

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+            if (text[i] == 0)
+                return 0;
+        }
+
+        const std::string materialName(text, materialNameLen);
+        text += materialNameLen;
+
+        auto* material = Game::DB_FindXAssetHeader(Game::XAssetType::ASSET_TYPE_MATERIAL, materialName.data()).material;
+        if (material == nullptr || material->techniqueSet == nullptr || material->techniqueSet->name == nullptr || strcmp(material->techniqueSet->name, "2d") != 0)
+            material = Game::DB_FindXAssetHeader(Game::XAssetType::ASSET_TYPE_MATERIAL, "default").material;
```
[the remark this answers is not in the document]
Laupetin: not a 3D material

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+                count++;
+                maxLengthRemaining--;
+            }
+
+            if(renderFlags & Game::TEXT_RENDERFLAG_CURSOR && count == cursorPos)
+            {
+                RotateXY(cosAngle, sinAngle, startX, startY, xa, xy, &xRot, &yRot);
+                Game::RB_DrawCursor(material, cursorLetter, xRot, yRot, sinAngle, cosAngle, font, xScale, yScale, color.packed);
+            }
+        }
```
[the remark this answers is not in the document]
Laupetin: Well this function was hooked to a very big extend before. I was considering doing this with hooks but at some point most of the function is hooked anyway. In that scenario i like to just reimplement the function. It also makes it easier to understand because this does not have compiler optimizations and other madness.
All in all i consider this better understandable than adding more and more hooks to it.
Reimplementing this also made me able to get rid of the hooks that were placed in the past so it also gets rid of some complexity from before.
There's not really a way to make it simple and i would personally consider this the best way (that's why i implemented it like this)

src/Components/Modules/TextRenderer.cpp
```diff
@@ -0,0 +1,1493 @@
...
+    }
+
+    int TextRenderer::R_TextWidth_Hk(const char* text, int maxChars, Game::Font_s* font)
+    {
+        auto lineWidth = 0;
+        auto maxWidth = 0;
+
+        if (maxChars <= 0)
+            maxChars = 0x7FFFFFFF;
```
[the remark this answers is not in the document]
Laupetin: Because the game uses an int for it and the original value is passed from the game. I'd personally use an unsigned type here as well but i prefer consistency with the game if the code interacts with it

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,214 @@
...
+			0.8f,
+			1.0f
+		};
+		static constexpr float HINT_COLOR[4]
+		{
+			0.6f,
+			0.6f,
+			0.6f,
+			1.0f
+		};
```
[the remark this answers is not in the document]
Laupetin: The colors are either made up by me or the game applies them in code instead of having them as data. So not possible sorry.

Laupetin:
Addressed all conversation that i think needed changes with exception of the localization ones. Will do them tomorrow, too exhausted this evening to do them now.

src/Components/Modules/TextRenderer.cpp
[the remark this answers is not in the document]
Rackover: Do we need that autocomplete in the console? I feel like this is mostly good for the chat.
I'm suggesting ENTER because that's what discord does, for example, and you used : to indicate emoji start/end similarly to what discord does - so I'd be enclined to make it behave in a similar way !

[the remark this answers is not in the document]
Rackover: We can pursue this conversation on discord if that's easier for you btw :D

momo5502:
Ngl, this PR is nice 😂

src/Components/Modules/PlayerName.cpp
momo5502: strncpy_s is just as ridiculous as fopen_s, stdncpy is fine

src/Components/Modules/TextRenderer.hpp
```diff
@@ -0,0 +1,220 @@
+#pragma once
+
+namespace Components
+{
+	enum TextColor
```
momo5502, line 5: Great idea of introducing an enum for that

diamante0018 (approved):

diamante0018 (dismissed):

Rackover (approved):

PR #134, opened 2021-10-04
[Dvar] Reset dvars changed by server when disconnecting

diamante0018:
The goal is to reset all archived dvars that are changed by the remote server after the client left the server.
This reasoning behind this pr is that servers may change Dvars such as "team name", player name, fov or the colour Dvars that controls the low ammo warning (just to name a few) and then forget to change them back to their original value.
As servers can modify any client Dvar that isn't read-only/write-protected (I think), the user should want the modified Dvars put back to their original value.

## Images
[1](https://i.imgur.com/2BC22zS.png)
[2](https://i.imgur.com/FFQAPdU.png)


Rackover (approved):
Looks okay to me.  For extra polish we could have the archivedvars.cfg check for duplicate variables being set consecutively (for example what if a server sets your name every round) and trim duplicate setas, but it's not a must-have.

diamante0018:
> Looks okay to me. For extra polish we could have the archivedvars.cfg check for duplicate variables being set consecutively (for example what if a server sets your name every round) and trim duplicate setas, but it's not a must-have.

I thought about it for a while and unfortunately, there is not an easy way to do this. Not even the game parser .cfg files. It just adds them wholly to the cbuf basically. So to check for duplicates I'd have to create a parser for the .cfg files. Given that normally only a few archive dvars should be modified so a few duplicates shouldn't cause any issue as the .cfg is executed after the player leaves the server. The only thing that could be done in my opinion, to keep IO overhead to a minimum, is to parse the .cfg before is executed and remove the duplicates so that the seta command gets executed once per dvar. but then again I don't think it's worth going through the trouble.

PR #141, opened 2021-11-07
[Movement] Add customizable speed scale for stance

diamante0018:
- This PR adds two new Dvars. The player_lastStandCrawlSpeedScale Dvar was already present in the game for debugging purposes. I originally wrote this code for IW5 and since the folks at pluto liked the idea I think it will be part of their client in some next future release. I ported the code to IW4 as well because I think this kind of customization can be beneficial to the client. I thoroughly tested the code on IW5 and I briefly tested it on IW4 because the code is identical.

- The new Dvars will affect how fast the player moves when he is in the prone stance, ducked stance or in the last stand stance. A single Dvar only affects the speed of the player when he is in that specific stance. For example, the player walking speed will be unaffected by changes to these Dvars. Changing the value of player_lastStandCrawlSpeedScale will not affect the player speed when he is prone (vice versa with changes to player_proneSpeedScale).

- I reversed PM_CmdScaleForStance and I made sure that it practically does the same thing as the original game function so that it doesn't break normal game behaviour. Instead of returning hard-coded 0.65f and 0.15f values for ducked and prone stance, it will return the current value of the corresponding Dvars.

- Rekitnator helped me re-write some code spat out by the disassembler that I could not make sense of.

# Note on the Dvar flag
I hook Dvar_RegisterFloat for the Dvar `player_lastStandCrawlSpeedScale` instead of manually setting the flags in memory. This is usually how Dvars are overwritten on s1x/iw6x. This allows me to get the Dvar pointer needed in the Movement module and to change the flags. I only set `DVAR_FLAG_REPLICATED` and `DVAR_FLAG_CHEAT` because I don't know what `DVAR_FLAG_UNKNOWN80` does. I think it is okay to not set the 0x80 flag because the Dvar is only read from in my hook (or that original game function) and is only supposed to be changed via a game script or through the console. That is why the cheat flag is needed. So yeah I don't see why I should leave the 0x80 flag because the purpose of this PR is to allow server owners to change the value of `player_lastStandCrawlSpeedScale`. It should not have any undocumented behaviour. If the flag ever gets documented so we know what it does then we can add it back to the Dvar if it does make sense to do so.

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -299,6 +299,9 @@ namespace Components
 		// Uncheat ui_debugMode
 		Utils::Hook::Xor<BYTE>(0x6312DE, Game::dvar_flag::DVAR_FLAG_CHEAT);
 
+		// Remove unknown flag (to prevent unknown behaviour)
+		Utils::Hook::Xor<BYTE>(0x448B42, Game::dvar_flag::DVAR_FLAG_UNKNOWN80);
+
 		// Hook dvar 'name' registration
 		Utils::Hook(0x40531C, Dvar::RegisterName, HOOK_CALL).install()->quick();
 
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -299,9 +299,6 @@ namespace Components
 		// Uncheat ui_debugMode
 		Utils::Hook::Xor<BYTE>(0x6312DE, Game::dvar_flag::DVAR_FLAG_CHEAT);
 
-		// Remove unknown flag from player_lastStandCrawlSpeedScale (prevent unknown behaviour)
-		Utils::Hook::Xor<BYTE>(0x448B42, Game::dvar_flag::DVAR_FLAG_UNKNOWN80);
-
 		// Hook dvar 'name' registration
 		Utils::Hook(0x40531C, Dvar::RegisterName, HOOK_CALL).install()->quick();
 
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -299,7 +299,7 @@ namespace Components
 		// Uncheat ui_debugMode
 		Utils::Hook::Xor<BYTE>(0x6312DE, Game::dvar_flag::DVAR_FLAG_CHEAT);
 
-		// Remove unknown flag (to prevent unknown behaviour)
+		// Remove unknown flag (prevent unknown behaviour)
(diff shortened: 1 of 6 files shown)


src/Components/Modules/Dvar.cpp
```diff
@@ -299,6 +299,9 @@ namespace Components
 		// Uncheat ui_debugMode
 		Utils::Hook::Xor<BYTE>(0x6312DE, Game::dvar_flag::DVAR_FLAG_CHEAT);
 
+		// Remove unknown flag from player_lastStandCrawlSpeedScale (prevent unknown behaviour)
+		Utils::Hook::Xor<BYTE>(0x448B42, Game::dvar_flag::DVAR_FLAG_UNKNOWN80);
```
Rackover, line 303: Is this necessary / does it otherwise create problems that we know of and can verify?

src/Components/Modules/Movement.cpp
```diff
@@ -0,0 +1,151 @@
...
+            Movement::PlayerLastStandCrawlSpeedScale = Dvar::Register<float>("player_lastStandCrawlSpeedScale",
+                0.2f, 0.0f, 5.0f, Game::DVAR_FLAG_CHEAT | Game::DVAR_FLAG_REPLICATED,
+                "The scale applied to the player speed when crawling in last stand");
+
+            Movement::PlayerProneSpeedScale = Dvar::Register<float>("player_proneSpeedScale",
+                0.15f, 0.0f, 5.0f, Game::DVAR_FLAG_CHEAT | Game::DVAR_FLAG_REPLICATED,
+                "The scale applied to the player speed when crawling");
+        });
+
+        Utils::Hook(0x572F34, Movement::PMCmdScaleForStanceStub, HOOK_CALL).install()->quick();
```
Rackover, line 144: If you can add a small comment on the right that says in which function this hook is (here it's in  PM_CmdScale_Walk I believe, the other one I didn't check) that would be very neat :)

Rackover (approved):
Tested, it's working, good to go as far as I'm concerned

src/Components/Modules/Dvar.cpp
diamante0018: No, it does not create any problems. But I thought about it I think I should hook the place where the dvar gets registered instead of doing this.  I will address in the next commit.

src/Components/Modules/Movement.cpp
diamante0018: Roger

momo5502 (approved):

PR #142, opened 2021-11-08
Better r draw debug

Rackover:


src/Components/Modules/Renderer.cpp
```diff
@@ -8,6 +8,18 @@ namespace Components
...
+	Dvar::Var Renderer::DrawSceneModelCollisions;
+	Dvar::Var Renderer::DrawModelBoundingBoxes;
+	Dvar::Var Renderer::DrawModelNames;
+	Dvar::Var Renderer::DrawAABBTrees;
+
+	float sceneModelsColor[4] = { 1.0f, 1.0f, 0.0f, 1.0f };
+	float dobjsColor[4] = { 0.0f, 1.0f, 1.0f, 1.0f };
+	float staticModelsColor[4] = { 1.0f, 0.0f, 1.0f, 1.0f };
+	float gentitiesColor[4] = { 1.0f, 0.5f, 0.5f, 1.0f };
+	int drawDistance = 80000;
```
diamante0018, line 21: I think this could be a constexpr since it is only read from.

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
+
+				case Game::ENT_HANDLER_TRIGGER_MULTIPLE:
+					if (ent->spawnflags & 0x40)
+						Game::R_AddDebugBounds(once, &b);
+					else
+						Game::R_AddDebugBounds(multiple, &b);
+					break;
+
+				default:
+					float rv = std::min((float)ent->handler, (float)5) / 5;
```
diamante0018, line 226: While you have a chance I would improve this code with something like
`float rv = std::min(static_cast<float>(ent->handler), 5.0f) / 5.0f;` and so on for the following expressions.

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
+					}
+				}
+			}
+			break;
+		}
+	}
+
+	void Renderer::DebugDrawModelNames()
+	{
+		auto val = DrawModelNames.get<Game::dvar_t*>()->current.integer;
```
diamante0018, line 356: `DrawModelNames.get<int>()` Will do the same

src/Utils/Utils.cpp
```diff
@@ -149,4 +149,11 @@ namespace Utils
 	{
 		return !(base1 + len1 <= base2 || base2 + len2 <= base1);
 	}
+
+	float Vec3SqrDistance(float(*v1)[3], float(*v2)[3])
```
diamante0018, line 153: I don't like the use of the pow. I think it is unnecessary and very slow.
I recall Laupetin saying that manually multiplying two values together is faster than using pow to the power of 2. 
If the exponent was larger than 2 then pow would be correct.

I would write this expression like this, making the parameters constant is not crucial.
```
float Vec3SqrDistance(const float v1[3], const float v2[3])
{
      return (v1[0]-v2[0])*(v1[0]-v2[0]) + (v1[1]-v2[1])*(v1[1]-v2[1]) * (v1[2]-v2[2])*(v1[2]-v2[2]);
}
```

RektInator:
Start using auto in your code.

RektInator (changes requested):
Start using auto in your code.

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
+
+		if (!val) return;
+
+		int clientNum = Game::CG_GetClientNum();
+		if (!Game::CL_IsCgameInitialized() || clientNum >= 18 || clientNum < 0 || !Game::g_entities[clientNum].client) return;
+
+		Game::gentity_t* clientEntity = &Game::g_entities[clientNum];
+
+		float(*playerPosition)[3] = &clientEntity->r.currentOrigin;
```
RektInator, line 365: ???????

src/Components/Modules/Renderer.cpp
```diff
@@ -230,6 +513,24 @@ namespace Components
 
 		// End vid_restart
 		Utils::Hook(0x4CA3A7, Renderer::PostVidRestartStub, HOOK_CALL).install()->quick();
+
+		Dvar::OnInit([]
+		{
+			static std::vector < char* > values =
+			{
+				const_cast<char*>("Disabled"),
```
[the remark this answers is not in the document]
RektInator: Why are you allocating a vector for static data??????

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
-//
-// 				HRESULT res = buffer->LockRect(&lockedRect, NULL, D3DLOCK_READONLY);
-//
-//
-// 				buffer->UnlockRect();
-// 			}
-// 		});
-
-		// Log broken materials
+		Scheduler::OnFrame([]() {
```
RektInator, line 445: Wrong thread.

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
+		switch (val) {
+		case 1:
+			for (auto i = 0; i < scene->sceneModelCount; i++)
+			{
+				if (!scene->sceneModel[i].model)
+					continue;
+
+				if (Utils::Vec3SqrDistance(playerPosition, &scene->sceneModel[i].placement.base.origin) < drawDistance)
+				{
+					auto b = scene->sceneModel[i].model->bounds;
```
RektInator, line 293: Inconsistency is key.

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +178,310 @@ namespace Components
...
+		}
+	}
+
+	void Renderer::DebugDrawSceneModelCollisions()
+	{
+		if (!DrawSceneModelCollisions.get<bool>()) return;
+
+		float green[4] = { 0.0f, 1.0f, 0.0f, 1.0f };
+
+		auto* scene = Game::scene;
```
RektInator, line 245: WHY

RektInator (dismissed):

mjkzy:
Um

src/Components/Modules/Renderer.cpp
Rackover: Booleans are not great for storing a number from 0 to 4

src/Components/Modules/Renderer.cpp
```diff
@@ -516,20 +531,21 @@ namespace Components
 
 		Dvar::OnInit([]
 		{
-			static std::vector < char* > values =
+			char* values[5] =
```
diamante0018, line 534: The array should be static. That is how the game does it.

src/Components/Modules/Renderer.cpp
```diff
@@ -266,12 +270,19 @@ namespace Components
...
 
 		if (!val) return;
 
+		// Ingame only
 		int clientNum = Game::CG_GetClientNum();
-		if (!Game::CL_IsCgameInitialized() || clientNum >= 18 || clientNum < 0 || !Game::g_entities[clientNum].client) return;
+		if (!Game::CL_IsCgameInitialized() ||
+			clientNum >= 18 ||
+			clientNum < 0 ||
+			!Game::g_entities[clientNum].client) {
```
diamante0018, line 282: Perhaps this line could make use of the nullptr. Not really sure on this one

src/Components/Modules/Renderer.cpp
```diff
@@ -353,12 +364,19 @@ namespace Components
...
```
diamante0018, line 376: Perhaps this line could make use of the nullptr. Not really sure on this one

src/Components/Modules/Renderer.cpp
```diff
@@ -223,9 +229,9 @@ namespace Components
 					break;
 
 				default:
-					float rv = std::min((float)ent->handler, (float)5) / 5;
-					float gv = std::clamp((float)ent->handler - 5, (float)0, (float)5) / 5;
-					float bv = std::clamp((float)ent->handler - 10, (float)0, (float)5) / 5;
+					auto rv = std::min(static_cast<float>(ent->handler), 5.0f) / 5.0f;
+					auto gv = std::clamp(static_cast<float>(ent->handler - 5), 0.f, 5.f) / 5.f;
```
diamante0018, line 233: I am not sure if you meant to type `5.f` because in the line above you typed `5.0f`

diamante0018 (changes requested):

diamante0018 (dismissed):

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +190,320 @@ namespace Components
...
+					Game::R_AddDebugBounds(color, &b);
+					break;
+				}
+			}
+		}
+	}
+
+	void Renderer::DebugDrawSceneModelCollisions()
+	{
+		if (!r_drawSceneModelCollisions.get<bool>()) return;
```
Laupetin, line 247: This is registered as an enum but only used as a bool?

Laupetin: Also i cant seem to make this do anything. Does this work?

src/Components/Modules/Renderer.cpp
```diff
@@ -230,6 +535,25 @@ namespace Components
...
+				"Scene Models",
+				"Scene Dynamic Objects",
+				"GfxWorld Static Models",
+				nullptr
+			};
+
+			Renderer::r_drawModelBoundingBoxes = Game::Dvar_RegisterEnum("r_drawModelBoundingBoxes", values, 0, Game::DVAR_FLAG_CHEAT, "Draw scene model bounding boxes");
+			Renderer::r_drawSceneModelCollisions = Game::Dvar_RegisterEnum("r_drawSceneModelCollisions", values, 0, Game::DVAR_FLAG_CHEAT, "Draw scene model collisions");
+			Renderer::r_drawTriggers = Game::Dvar_RegisterBool("r_drawTriggers", false, Game::DVAR_FLAG_CHEAT, "Draw triggers");
+			Renderer::r_drawModelNames = Game::Dvar_RegisterEnum("r_drawModelNames", values, 0, Game::DVAR_FLAG_CHEAT, "Draw all model names");
```
Laupetin, line 553: Having this separate from r_drawModelBoundingBoxes with also an enum is kind of weird for me. I would have expected a bool for drawModelNames and the thing it draws being the same as r_drawModelBoundingBoxes

src/Components/Modules/Renderer.cpp
```diff
@@ -166,39 +190,320 @@ namespace Components
...
+			break;
+		}
+	}
+
+	void Renderer::DebugDrawAABBTrees()
+	{
+		if (!r_drawAABBTrees.get<bool>()) return;
+
+		Game::clipMap_t* clipMap = *reinterpret_cast<Game::clipMap_t**>(0x7998E0);
+		//Game::GfxWorld* gameWorld = *reinterpret_cast<Game::GfxWorld**>(0x66DEE94);
```
Laupetin, line 449: commented out code?

src/Components/Modules/Renderer.cpp
Rackover: It does, but not all maps have scene models, maybe it was the map?   i'm fixing the dvar type regardless, it might have been that

src/Components/Modules/Renderer.cpp
Rackover: It's a valid point, i'm gonna think about it a little

Rackover: I think I want to keep them separate for now because I use them for different purposes. Boundaries for collisions/boundaries/xmodel bugs and names to know the name of an xmodel in the world. Some areas have a lot of models and displaying both always could clutter the screen with information I don't want. So I think I'm going to keep it this way if that's ok for you 🙂

Laupetin: Alright

Laupetin (approved):

PR #143, opened 2021-11-10
[QuickPatch] Quick Refactor

diamante0018:
In light of a recent pull request, it came to my attention that some things were done incorrectly in the QuickPatch module.
This pr should be beneficial because:
- A vector is no longer being allocated when registering an enum dvar
- Added comments to InvalidNameCheckStub
- Replace the explicit call to strncpy with a call to actual game function. It was argued before that it is the best approach. (https://github.com/XLabsProject/iw4x-client/pull/92#discussion_r663989104).
- Proper use of const in InvalidNameCheck
- Use the dvar class wrapper when appropriate

### DIFF
--- a/src/Components/Modules/QuickPatch.cpp
+++ b/src/Components/Modules/QuickPatch.cpp
@@ -3,6 +3,7 @@
 namespace Components
 {
 	int QuickPatch::FrameTime = 0;
+	Dvar::Var QuickPatch::r_customAspectRatio;
 
 	void QuickPatch::UnlockStats()
 	{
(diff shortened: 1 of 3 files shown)


Rackover (approved):

src/Components/Modules/QuickPatch.cpp
```diff
@@ -268,30 +267,31 @@ namespace Components
 		}
 	}
 
-	Game::dvar_t* QuickPatch::r_customAspectRatio;
 	Game::dvar_t* QuickPatch::Dvar_RegisterAspectRatioDvar(const char* name, char**, int defaultVal, int flags, const char* description)
 	{
-		static std::vector < char * > values =
+		static char* r_aspectRatioEnum[] =
```
momo5502, line 272: Uh. That's bad

momo5502: Adjusting DvarRegisterEnum to accept const char** would be better

momo5502 (approved):

src/Components/Modules/QuickPatch.cpp
diamante0018: Done

Dss0 (approved):

PR #145, opened 2021-11-16
[Logger] Use correct enum over numbers for com_error

diamante0018:
Should be an improvement as it's easier to use an enum over numbers that you have to remember


Rackover (approved):
Good to go

PR #148, opened 2021-11-29
[Feature] Add options to disable elevators

diamante0018:
# About
A stub is added inside PM_GroundTrace, it will allow skipping the only call to PM_CorrectAllSolidStub (if the sv_disableElevators Dvar is set to 1) and will disable elevators. This feature should be welcomed by server owners that don't want people to glitch out of maps.

- sv_disableElevators: default to false so elevators are only disabled if it's set to 1 by the server owner. Dvar is cheat protected and replicated as it should be imo

sv_disableElevators follows the principles described here: [Research](https://xoxor4d.github.io/research/mw2-elevator/).
To recap, xoxor4d observes that skipping a call to PM_CorrectAllSolid or PM_JitterPoint in later Cods disables elevators.

# Credits
 * [xoxor4d](https://github.com/xoxor4d)
 - Matrix for general feedback

### DIFF
--- a/src/Components/Modules/Elevators.cpp
+++ b/src/Components/Modules/Elevators.cpp
@@ -1,0 +1,81 @@
+#include "STDInclude.hpp"
+
+namespace Components
+{
+	Game::dvar_t* Elevators::SV_EnableEasyElevators;
+
+	__declspec(naked) void Elevators::PM_CorrectAllSolidStub()
+	{
+		__asm
+		{
+			push ecx
+			mov ecx, Elevators::SV_EnableEasyElevators
+			cmp byte ptr [ecx + 16], 1
+			pop ecx
+
+			// Always elevate if SV_EnableEasyElevators is set to 1
+			je elevate
+
+			// Original code
+			cmp byte ptr [eax + 0x29], 0
+
+			// Original code flow
+			jz elevate
+
+			push 0x5734FF
+			retn
+
+		elevate:
+			push 0x57353D
+			retn
+		}
+	}
+
+	__declspec(naked) void Elevators::PM_CheckDuckStub()
+	{
+		__asm
+		{
+			push eax
+			mov eax, Elevators::SV_EnableEasyElevators
+			cmp byte ptr [eax + 16], 1
+			pop eax
+
+			// Always stand if SV_EnableEasyElevators is set to 1
+			je stand
+
+			// Original code
+			cmp byte ptr [esp + 0x38], 0
+
+			// Original code flow
+			jnz noStand
+
+		stand:
+
+			push 0x570ED4
+			retn
+
+		noStand:
+			push 0x570EF4
+			retn
+		}
+	}
+
+	Elevators::Elevators()
+	{
+		Dvar::OnInit([]
+		{
+			Elevators::SV_EnableEasyElevators = Game::Dvar_RegisterBool("sv_enableEasyElevators",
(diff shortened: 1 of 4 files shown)


src/Components/Modules/Elevators.cpp
```diff
@@ -0,0 +1,81 @@
...
+	{
+		Dvar::OnInit([]
+		{
+			Elevators::SV_EnableEasyElevators = Game::Dvar_RegisterBool("sv_enableEasyElevators",
+				false, Game::DVAR_FLAG_CHEAT | Game::DVAR_FLAG_REPLICATED,
+				"Enable easy elevators for trickshotting");
+		});
+
+		Utils::Hook(0x5734F9, Elevators::PM_CorrectAllSolidStub, HOOK_JUMP).install()->quick();
+		Utils::Hook::Nop(0x5734FE, 1);
```
Rackover, line 73: Comment here please :)

src/Components/Modules/Elevators.cpp
```diff
@@ -0,0 +1,81 @@
...
+			// Original code flow
+			jnz noStand
+
+		stand:
+
+			push 0x570ED4
+			retn
+
+		noStand:
+			push 0x570EF4
```
Rackover, line 58: The four memory addresses written as raw values in this and the other stub should to be commented as to tell what they point to !

Rackover (commented):
Looks good but I have not tested it

src/Components/Modules/Elevators.cpp
diamante0018: They don't point to anything. All I am doing is exiting the function and jumping to that address in memory. 
My comments will be brief.

src/Components/Modules/Elevators.cpp
diamante0018: Done

skiff:
I like the disabling elevators, giving server owners the ability to patch a glitch in the game not everyone likes is generally a good idea. I don't think the dvar should be cheat protected though. I think the flags should match the `sv_enableBounces` dvar

diamante0018:
> I like the disabling elevators, giving server owners the ability to patch a glitch in the game not everyone likes is generally a good idea. I don't think the dvar should be cheat protected though. I think the flags should match the `sv_enableBounces` dvar

It will be done. Only replicated it is.

momo5502 (approved):

src/Components/Modules/Elevators.cpp
```diff
@@ -0,0 +1,55 @@
...
+			mov eax, Elevators::SV_DisableElevators
+			cmp byte ptr [eax + 16], 1
+			pop eax
+
+			// Always skip PM_CorrectAllSolid if SV_DisableElevators is set to 1
+			je noElevators
+
+			// Original code
+			cmp byte ptr [esp + 0x50], 0
+			rep movsd
```
momo5502, line 21: Code looks right to me, but always make sure to check which registers rep, especially rep movsd touches, just to be sure you don't clobber any. But it seems that you don't.

diamante0018: Yeah, I've checked again now and to the best of my knowledge, the rep instruction is using the zf flag here to perform the operation mvsd set by the cmp instruction anyway the only register I use is the eax register but that's protected by push pop done earlier. That is simply the original game code that my stub skip. so yeah all is good.

In addition following xoxor4d advice, I spent some time earlier playing with this patch on to see if anything unintended happens but it didn't so all should be good.

PR #157, opened 2021-12-28
[Elevators] Add prone elevators

diamante0018:
Add prone elevators to the easy settings.

If elevators are set to easy the new hooked function will make sure the ducked + prone flags are always removed from player state


Laupetin:
Works fine, looks fine. However Prone into Crouch still says it's blocked. Maybe that's something you still wanna do.

diamante0018:
> Works fine, looks fine. However Prone into Crouch still says it's blocked. Maybe that's something you still wanna do.

Added

Laupetin (approved):

PR #158, opened 2021-12-28
[Movement] Refactor+ features

diamante0018:
Add ufo+noclip speed scaler.
Refactor other functions present in the module as when I initially coded them I forgot a few minor details in my opinion.
Refactor bounces so it includes double bounces and bounce from all angles.

Probably need some feedback on the naming conventions of the two new dvars

## Xoxor4d
This pr would not be possible without his knowledge of the engine that he made publicly available.


diamante0018:
One section of the code as been reviewed by momo as I ported it to iw6x (bounce/bounce from all angles)
The new collision/ejection hooks are made without using a single line of assembly like iw6x.
the other features have been tested by people are than me so we know they work as intended.
As said by the pr description, most of the concepts behind these engine modifications have been validated by xoxor4d.
This is ready to be merged

Dss0 (approved):

PR #159, opened 2022-01-01
[Chat] Mute player (prevents them from using the chat)

diamante0018:
This is a feature request by a staff member.

So allow me to briefly go other the design choice I made and why. This is a command that will most likely be used by IW4M via rcon. So I took into consideration that IW4M can manage the duration of a "mute" penalty on its own using a plugin.
Also if a player disconnects it can still be removed from the mute list. How? I used the GUID (named SteamId in the code) which is known to IW4M since it can parse it from the status command reply.

I decided to use the SteamID as a client identifier because IW4M can then unmute players even if they are offline. So mute penalties can stay persistent even if a player disconnects and then be lifted automatically. Imagine a situation where a player leaves the server and reconnects in an attempt to evade the ban (It's what happens in 4d1 client clones where they have InfinityScript), that should be taken care of.

Examples:
`muteClient <client number>` Adds the player to the mute list
`unmute <client number or the GUID>` or `unmute all`. Removes all clients or the specified client from the mute list.

### DIFF
--- a/src/Components/Modules/Chat.cpp
+++ b/src/Components/Modules/Chat.cpp
@@ -8,10 +8,13 @@ namespace Components
 
 	bool Chat::SendChat;
 
+	std::mutex Chat::AccessMutex;
 	std::unordered_set<std::uint64_t> Chat::MuteList;
 
 	const char* Chat::EvaluateSay(char* text, Game::gentity_t* player)
 	{
+		std::lock_guard<std::mutex> _(Chat::AccessMutex);
+
 		Chat::SendChat = true;
 
 		if (text[1] == '/')
@@ -205,12 +208,14 @@ namespace Components
 
 	void Chat::MuteClient(const Game::client_t* client)
 	{
+		std::lock_guard<std::mutex> _(Chat::AccessMutex);
+
 		if (Chat::MuteList.find(client->steamID) == Chat::MuteList.end())
 		{
 			Chat::MuteList.insert(client->steamID);
 
 			Logger::Print("%s was muted\n", client->name);
-			Game::SV_GameSendServerCommand(Bots::GetClientNum(client), 0,
+			Game::SV_GameSendServerCommand(client->gentity->s.number, 0,
 				Utils::String::VA("%c \"You were muted\"", 0x65));
 			return;
 		}
@@ -222,11 +227,21 @@ namespace Components
 
 	void Chat::UnmuteClient(const Game::client_t* client)
 	{
-		Chat::MuteList.erase(client->steamID);
+		Chat::UnmuteInternal(client->steamID);
 
 		Logger::Print("%s was unmuted\n", client->name);
-		Game::SV_GameSendServerCommand(Bots::GetClientNum(client), 0,
+		Game::SV_GameSendServerCommand(client->gentity->s.number, 0,
 			Utils::String::VA("%c \"You were unmuted\"", 0x65));
+	}
+
+	void Chat::UnmuteInternal(const std::uint64_t id, bool everyone)
+	{
+		std::lock_guard<std::mutex> _(Chat::AccessMutex);
+
+		if (everyone)
+			Chat::MuteList.clear();
+		else
+			Chat::MuteList.erase(id);
 	}
 
 	void Chat::AddChatCommands()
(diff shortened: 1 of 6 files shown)


src/Components/Modules/Chat.cpp
```diff
@@ -8,17 +8,26 @@ namespace Components
 
 	bool Chat::SendChat;
 
+	std::unordered_set<std::uint64_t> Chat::MuteList;
```
momo5502, line 11: You should guard this variable with a mutex. It can be read/written to concurrently by different threads

src/Components/Modules/Chat.cpp
```diff
@@ -8,17 +8,26 @@ namespace Components
...
 
 		if (text[1] == '/')
 		{
-			SendChat = false;
+			Chat::SendChat = false;
 			text[1] = text[0];
 			++text;
 		}
 
+		if (Chat::MuteList.find(Game::svs_clients[player->s.number].steamID) != Chat::MuteList.end())
```
momo5502, line 24: Mutex

src/Components/Modules/Chat.cpp
```diff
@@ -194,9 +203,99 @@ namespace Components
 		}
 	}
 
+	void Chat::MuteClient(const Game::client_t* client)
+	{
+		if (Chat::MuteList.find(client->steamID) == Chat::MuteList.end())
```
momo5502, line 208: Mutex

src/Components/Modules/Chat.cpp
```diff
@@ -194,9 +203,99 @@ namespace Components
...
+		}
+
+		Logger::Print("%s is already muted\n", client->name);
+		Game::SV_GameSendServerCommand(-1, 0,
+			Utils::String::VA("%c \"%s is already muted\"", 0x65, client->name));
+	}
+
+	void Chat::UnmuteClient(const Game::client_t* client)
+	{
+		Chat::MuteList.erase(client->steamID);
```
momo5502, line 225: Mutex

momo5502 (changes requested):

src/Components/Modules/Chat.cpp
diamante0018: Thanks for the review. As I addressed it, I created a helper function to guard the list when it comes to the unmute command. Now all access is guarded by a mutex.
If you don't mind, please make sure the way I went about addressing your review is correct. Thanks.

diamante0018:
Hi @momo5502 I addressed your review. Could you please make sure I got it right?
Thanks

src/Components/Modules/Chat.cpp
```diff
@@ -8,17 +8,29 @@ namespace Components
 
 	bool Chat::SendChat;
 
+	std::mutex Chat::AccessMutex;
+	std::unordered_set<std::uint64_t> Chat::MuteList;
+
 	const char* Chat::EvaluateSay(char* text, Game::gentity_t* player)
 	{
-		SendChat = true;
+		std::lock_guard<std::mutex> _(Chat::AccessMutex);
```
momo5502, line 16: You're locking this for the entire function. Even though this is not necessary

src/Components/Modules/Chat.cpp
```diff
@@ -8,17 +8,29 @@ namespace Components
...
 
 		if (text[1] == '/')
 		{
-			SendChat = false;
+			Chat::SendChat = false;
 			text[1] = text[0];
 			++text;
 		}
 
+		if (Chat::MuteList.find(Game::svs_clients[player->s.number].steamID) != Chat::MuteList.end())
```
momo5502, line 27: I would use a unique lock before the if and unlock the lock inside the if and also outside, that way the lock is not helf for longer than needed, but also raii guarded

src/Components/Modules/Chat.cpp
```diff
@@ -194,9 +206,111 @@ namespace Components
 		}
 	}
 
+	void Chat::MuteClient(const Game::client_t* client)
+	{
+		std::lock_guard<std::mutex> _(Chat::AccessMutex);
+
+		if (Chat::MuteList.find(client->steamID) == Chat::MuteList.end())
```
momo5502, line 213: Same here, unique lock and manual unlock as soon as not needed anymore

momo5502 (approved):

PR #178, opened 2022-01-27
README + Cleanup project folder

diamante0018:


README.md
```diff
@@ -36,21 +36,20 @@
...
-| `-reallybigdumps`           | Unused.                                        |
-| `-bigminidumps`             | Mini dumps.                                    |
-| `-reallybigminidumps`       | Big mini dumps.                                |
-| `-dump`                     | Prints asset info to a .ents file.             |
-| `-monitor`                  | Enables monitor.                               |
-| `-nointro`                  | Skips game's intro.                            |
-| `-version`                  | Prints IW4X version.                           |
-| `-zonebuilder`              | Enables zone builder.                          |
-| `-nosteam`                  | Disables Steam features.                       |
+| `-entries`                  | Print ff info after being loaded.              |
```
Rackover, line 39: Prints to the console a list of every asset as they are loaded from zonefiles

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-stdout`                   | Redirect stdout to the console.                |
```
Rackover, line 40: Redirects all logging output to the terminal iw4x is started from, or if there is none, creates a new terminal window to write log information in

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-console`                  | Create game's external console.                |
```
Rackover, line 41: Allows the game to display its own separate interactive console window

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-dedicated`                | Enabled dedicated server.                      |
```
Rackover, line 42: Starts the game as a headless dedicated server

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-scriptablehttp`           | Enable HTTP gsc functions.                     |
+| `-bigdumps`                 | Include all code sections from loaded modules. |
+| `-bigminidumps`             | Include all code sections from loaded modules. |
+| `-reallybigminidumps`       | Include data sections from all loaded modules. |
+| `-dump`                     | Write info of loaded assets to the raw folder. |
+| `-monitor`                  | Indicated whenever the game console is active. |
+| `-nointro`                  | Skip game's cinematic intro.                   |
+| `-version`                  | Print IW4x build info                          |
+| `-zonebuilder`              | Enable zone builder.                           |
```
Rackover, line 51: Starts the interactive zonebuilder tool console instead of starting the game

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-entries`                  | Print ff info after being loaded.              |
+| `-stdout`                   | Redirect stdout to the console.                |
+| `-console`                  | Create game's external console.                |
+| `-dedicated`                | Enabled dedicated server.                      |
+| `-scriptablehttp`           | Enable HTTP gsc functions.                     |
+| `-bigdumps`                 | Include all code sections from loaded modules. |
+| `-bigminidumps`             | Include all code sections from loaded modules. |
+| `-reallybigminidumps`       | Include data sections from all loaded modules. |
+| `-dump`                     | Write info of loaded assets to the raw folder. |
+| `-monitor`                  | Indicated whenever the game console is active. |
```
Rackover, line 48: I don't understand what this means and I don't understand what the previous comment meant either

README.md
```diff
@@ -36,21 +36,20 @@
...
+| `-nointro`                  | Skip game's cinematic intro.                   |
+| `-version`                  | Print IW4x build info                          |
+| `-zonebuilder`              | Enable zone builder.                           |
+| `-nosteam`                  | Same effect as enabling cl_anonymous dvar.     |
```
Rackover, line 52: Which is ? :/

This is particularly unhelpful, it's even less clear than the previous comment there was here. Do not be afraid to write as many text as you need, you are not limited in space, Markdown will make it all fit in the table cell regardless

Rackover (changes requested):
Most descriptions are less clear now than they even were before the rewrite 😰   
You need to write as many text as needed for each and be descriptive when necessary, it doesn't have to look good in ascii format because this is all markdown that will get automatically formatted by github. Add as much text to each cell as you should. Do not limit yourself to ascii art

README.md
diamante0018: It appears as it's a flag for internal use. Like to know whether the client is headless or not. Not so sure either

Rackover (dismissed):
I guess it's okay but I would prefer someone else to weigh in

Dss0 (approved):

PR #179, opened 2022-01-27
[IFont] Use Zone allocator instead of static mem allocator

diamante0018:
Needs testers

### DIFF
--- a/src/Components/Modules/AssetInterfaces/IFont_s.cpp
+++ b/src/Components/Modules/AssetInterfaces/IFont_s.cpp
@@ -53,10 +53,10 @@ namespace Assets
 				auto& glyph = glyphs[i++];
 
 				glyph.letter = ch;
-				glyph.s0 = static_cast<float>(x / pw);
-				glyph.s1 = static_cast<float>((x + gw) / pw);
-				glyph.t0 = static_cast<float>(y / ph);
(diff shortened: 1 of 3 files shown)


src/Components/Modules/AssetInterfaces/IFont_s.cpp
```diff
@@ -53,15 +53,15 @@ namespace Assets
...
 				glyph.letter = ch;
-				glyph.s0 = x / static_cast<float>(pw);
-				glyph.s1 = (x + gw) / static_cast<float>(pw);
-				glyph.t0 = y / static_cast<float>(ph);
-				glyph.t1 = (y + gh) / static_cast<float>(ph);
+				glyph.s0 = static_cast<float>(x / pw);
+				glyph.s1 = static_cast<float>((x + gw) / pw);
+				glyph.t0 = static_cast<float>(y / ph);
+				glyph.t1 = static_cast<float>((y + gh) / ph);
 				glyph.pixelWidth = static_cast<char>(gw);
```
GEEKiDoS, line 60: Wouldn't these changes will case the decimal part of number to be cutted? they are int

diamante0018: I see. I must have misunderstood the purpose of this section. I reverted the changes.

Dss0 (approved):

PR #182, opened 2022-01-30
[Map] Fix ent string

diamante0018:


Rackover:
Tested , loads maps correctly as far as I've tested it, but I'm concerned with a crash on overgrown that I could not (?) reproduce

diamante0018:
Ready to be re tested to confirm it fixed crash

diamante0018:
Thoroughly tested by me, can be merged as it fixes the crash.

Dss0 (approved):

PR #183, opened 2022-02-03
Fix gamepad vars

diamante0018:
Petition to replace
`aim_turnrate_pitch_ads = Dvar::Var("aim_turnrate_pitch_ads");`
with
`aim_turnrate_pitch_ads = Dvar::Var(0x0);` <- replace with the actual address of where the pointer is stored


diamante0018:
Just for the record. This patch simply moves the hook so that the gamepad code executes after the dvars have been registered by the game. I was initially aiming to rework the dvar module but I prefer if that was a different pull request. I also corrected the function definition of a bunch of dvar related function.
It was theoretically correct to say that some of the function were returning a dvar pointer because the eax registered was left untouched by the game code but I know from very clean reverse engineering that it's not the case.

Dss0 (approved):

PR #184, opened 2022-02-03
[Dedicated] Fix questionable patch

diamante0018:
This is usually how s1x and iw6x override dvars


diamante0018:
As you can see from this screenshot the patch works fine (dedicate & zonetool builder only like it currently is)

[Img](https://i.imgur.com/eOrX0JW.pngl)

Dss0 (approved):

PR #186, opened 2022-02-07
Fix cbuf not being terminated

diamante0018:


src/Components/Modules/StartupMessages.cpp
```diff
@@ -28,7 +28,7 @@ namespace Components
 			Game::Dvar_SetStringByName("ui_startupMessage", message.data());
 			Game::Dvar_SetStringByName("ui_startupMessageTitle", Utils::String::VA("Messages (%d/%d)", StartupMessages::TotalMessages - StartupMessages::MessageList.size(), StartupMessages::TotalMessages));
 			Game::Dvar_SetStringByName("ui_startupNextButtonText", StartupMessages::MessageList.size() ? "Next" : "Close");
-			Game::Cbuf_AddText(0, "openmenu startup_messages");
+			Game::Cbuf_AddText(0, "openmenu startup_messages\n");
```
mjkzy, line 31: Why can't you just insert text for this too and remove the \n?

diamante0018: cbuf_addtext is faster in some contextes but inserttext is faster in others

PR #188, opened 2022-02-10
[Exception] Refactor component & copy exception message to clipboard

diamante0018:
Simply because I dislike manually typing the address from the message box. I'd like to have it in the clipboard so I can just copy-paste it into whatever tool I use for debugging the error.


diamante0018:
Tested and it works as intented.

momo5502 (approved):

PR #219, opened 2022-04-13
Add branding component

diamante0018:
- Remove random profanity in the premake5. 
- 3 new dvars that control "development" branding-
- Move iw4x in menu (UI) branding to the top right corner of the screen (Conflict with development branding which has always been bottom right corner)


diamante0018:
https://i.imgur.com/TLzOmH9.png

src/Components/Modules/Branding.cpp
```diff
@@ -0,0 +1,131 @@
...
+#endif
+
+		// IW4x is technically a beta
+		const auto* result = Utils::String::VA("%s %s build %s %s",
+			buildType, "(Beta)", Branding::GetBuildNumber(), reinterpret_cast<const char*>(0x7170A0));
+
+		return result;
+	}
+
+	void Branding::Dvar_SetVersionString(const Game::dvar_t* dvar, const char* /*value*/)
```
Laupetin, line 66: [[maybe_unused]] if you wanna use it :P

src/Components/Modules/Branding.cpp
```diff
@@ -0,0 +1,131 @@
...
+		constexpr auto maxChars = std::numeric_limits<int>::max();
+		// Default colours
+		constexpr Game::vec4_t shadowColor = {0.0f, 0.0f, 0.0f, 0.69f};
+		constexpr Game::vec4_t color = {0.4f, 0.69f, 1.0f, 0.69f};
+
+		auto* const placement = Game::ScrPlace_GetUnsafeFullPlacement();
+		auto* const font = Game::UI_GetFontHandle(placement, 0, 0.5f);
+
+		const auto width = Game::UI_TextWidth((*Version)->current.string, 0, font, fontScale);
+		const auto height = Game::UI_TextHeight(font, fontScale);
```
Laupetin, line 23: Convert these two into float before using them otherwise it's undefined whether float + int is float or int

Laupetin (changes requested):

src/Components/Modules/Branding.cpp
diamante0018: You caught me red handed!

src/Components/Modules/Branding.cpp
```diff
@@ -0,0 +1,122 @@
...
+	{
+		// Default values
+		constexpr auto fontScale = 0.25f;
+		constexpr auto maxChars = std::numeric_limits<int>::max();
+		// Default colours
+		constexpr Game::vec4_t shadowColor = {0.0f, 0.0f, 0.0f, 0.69f};
+		constexpr Game::vec4_t color = {0.4f, 0.69f, 1.0f, 0.69f};
+
+		auto* const placement = Game::ScrPlace_GetUnsafeFullPlacement();
+		auto* const font = Game::UI_GetFontHandle(placement, 0, 0.5f);
```
Laupetin, line 20: Is there a reason why this uses `0.5f` instead of `fontScale`?

src/Components/Modules/Branding.cpp
```diff
@@ -0,0 +1,122 @@
...
+	Game::dvar_t** Branding::Version = reinterpret_cast<Game::dvar_t**>(0x1AD7930);
+
+	void Branding::CG_DrawVersion()
+	{
+		// Default values
+		constexpr auto fontScale = 0.25f;
+		constexpr auto maxChars = std::numeric_limits<int>::max();
+		// Default colours
+		constexpr Game::vec4_t shadowColor = {0.0f, 0.0f, 0.0f, 0.69f};
+		constexpr Game::vec4_t color = {0.4f, 0.69f, 1.0f, 0.69f};
```
Laupetin, line 17: The debug text you replaced does tint this green when a debugger is attached. Not super important but was a nice touch maybe you wanna do something like that too? :)

Laupetin (approved):

src/Components/Modules/Branding.cpp
diamante0018: Will change

src/Components/Modules/Branding.cpp
diamante0018: Not really a fan of that

Laupetin (approved):

PR #225, opened 2022-04-15
[DRAFT] Moved essential code to run a single game into a separate component t…

Rackover:
…o allow for a minimal start

the weapon bit is kind of extra but iw4x files contain weapons in duplicate, so making sure they're only loaded once (unless the weapons component is loaded) is necessary for running proper.

The target we want the minimal component to be able to achieve is : Launch the game, get (alone) into a multiplayer map, play for as long as you would need to, with the smallest possible amount of modifications (no modifications that are not strictly necessary to run a vanilla map).

This restructuration's goal is to make possible to disable/enable components at will to facilitate debugging and still allow to run a vanilla map from the basegame with everything disabled 🙂

### DIFF
--- a/src/Components/Modules/Core.cpp
+++ b/src/Components/Modules/Core.cpp
@@ -2,9 +2,9 @@
 
 namespace Components
 {
-	int Core::weaponsLoaded = 0;
-	Game::XAssetEntry* Core::lastWeaponEntry = nullptr;
-	std::unordered_set<std::string> Core::skippedWeapons{};
+	const unsigned long DEFAULT_WEAPON_LIMIT = *reinterpret_cast<unsigned long*>(0x403783);
+
+	std::unordered_set<std::string> Core::loadedWeapons{};
 
 	Core::Core()
 	{
@@ -32,7 +32,7 @@ namespace Components
 		Utils::Hook::Set<BYTE>(0x51F59F, 0xEB);
 
 		// Asset limits - instead of erroring when going above, just ignore further loads
-		Utils::Hook(0x4A473C, Core::AddXAssetIfWeaponUnderLimit, HOOK_CALL).install()->quick(); 
+		Utils::Hook(0x4A473C, Core::AddUniqueWeaponXAsset, HOOK_CALL).install()->quick(); 
 
 		/////////////
 		/// various changes to SV_DirectConnect-y stuff to allow non-party joinees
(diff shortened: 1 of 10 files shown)


src/Components/Modules/Core.cpp
```diff
@@ -0,0 +1,83 @@
+#include <STDInclude.hpp>
+
+namespace Components
+{
+	const unsigned long DEFAULT_WEAPON_LIMIT = *reinterpret_cast<unsigned long*>(0x403783);
```
diamante0018, line 5: This is not good, we never had to read constants from the game memory used n the middle of a function.
constexpr auto please.

src/Components/Modules/Core.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Components
+{
+	class Core : public Component
+	{
+	public:
+		Core();
+
+	private:
+		static Game::XAssetEntry* __cdecl AddUniqueWeaponXAsset(Game::XAssetType type, Game::XAssetHeader* a2);
```
diamante0018, line 11: Remove __cdecl

diamante0018, line 11: Rename a2 to header

src/Components/Modules/Core.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+		// Bypass "user already registered"  (EXE_TRANSMITERROR)
+		Utils::Hook::Set<BYTE>(0x401C15, 0xEB);
+
+		// Make sure preDestroy is called when the game shuts down
+		Scheduler::OnShutdown(Loader::PreDestroy);
+	}
+
+	Game::XAssetEntry* __cdecl Core::AddUniqueWeaponXAsset(Game::XAssetType type, Game::XAssetHeader* asset) {
+		if (Loader::GetInstance<Weapon>() == nullptr && asset && asset->weapon) {
```
diamante0018: I prefer explicit != nullptr

src/Components/Modules/Core.cpp
```diff
@@ -0,0 +1,83 @@
...
+			std::string name = std::string(asset->weapon->szInternalName);
```
diamante0018, line 62: You have two options IMO:
`const std::string name(asset->weapon->szInternalName);`
`const std::string name = asset->weapon->szInternalName;`

diamante0018, line 62: How sure are you that szInternalName is never going to be a nullptr?
Gonna throw C++ exception if it is

diamante0018 (commented):
use data over .cstr. 
Remove unnecessary components namespace
allman braces

Rackover:
I'm not sure it's worth merging in at all, but it's still a very practical setup for debugging. what do ?

diamante0018:
> I'm not sure it's worth merging in at all, but it's still a very practical setup for debugging. what do ?

I'm in favour of removing "random" patches from here and there(mostly QuickPatch) and grouping them in one module.
I think we can merge this. I don't recall, this code works right?

Rackover:
> > I'm not sure it's worth merging in at all, but it's still a very practical setup for debugging. what do ?
> 
> I'm in favour of removing "random" patches from here and there(mostly QuickPatch) and grouping them in one module. I think we can merge this. I don't recall, this code works right?

It works but it moved some stuff out of "party" into Core, among other, and maybe that's not for the best? idk. But moving stuff out of quickpatch is generally good, except moving them to something called "core" is just the same as if we had been calling it "quickpatch2"

diamante0018:
I see that, maybe put back stuff into party and playlist component but leave stuff take out of quickpatch into core?
does that work for you?

Rackover:
No, if we do that core won't be different from just another quickpatch module... it doesn't make sense. The thing I tried to do was poorly thought through because if you want to have a minimal core component you need it to do several different things at once, which goes against the philosophy of having multiple different components one for each thing. iw4x needs a minimal option for easier debugging, but this ain't it !
Maybe have each component have a `Critical` part that gets hooked and everything always, and a `Non-Critical` part that (like wrapped with macros for instance?) that is simply ignored if you have a certain build configuration. Regardless, this PR is doing it wrong! 
Closing it

PR #230, opened 2022-04-29
Add RawMouse Compoment

GEEKiDoS:
And enable dpi awareness for game window.
Should improve mouse accuracy 

PS: should fixes mouse accelaction issue when mouse polling rate > 125Hz on iw6x/s1x (going to port this now

### DIFF
--- a/src/Components/Loader.hpp
+++ b/src/Components/Loader.hpp
@@ -137,3 +137,5 @@ namespace Components
 
 #include "Modules/Gamepad.hpp"
 #include "Modules/ScriptExtension.hpp"
+
+#include "Modules/RawMouse.hpp"
--- a/src/Components/Loader.hpp
+++ b/src/Components/Loader.hpp
@@ -134,8 +134,6 @@ namespace Components
 #include "Modules/Movement.hpp"
 #include "Modules/Elevators.hpp"
 #include "Modules/ClientCommand.hpp"
-
 #include "Modules/Gamepad.hpp"
 #include "Modules/ScriptExtension.hpp"
-
 #include "Modules/RawMouse.hpp"
--- a/src/Components/Modules/RawMouse.cpp
+++ b/src/Components/Modules/RawMouse.cpp
@@ -2,80 +2,164 @@
 
 namespace Components
 {
-	void IN_ClampMouseMove()
+	Dvar::Var RawMouse::m_rawinput;
+
+	void RawMouse::IN_ClampMouseMove()
 	{
-        tagRECT rc;
-        tagPOINT curPos;
+		tagRECT rc;
+		tagPOINT curPos;
 
-        GetCursorPos(&curPos);
-        GetWindowRect(Game::g_wv->hWnd, &rc);
-        bool isClamped = false;
-        if (curPos.x >= rc.left)
-        {
-            if (curPos.x >= rc.right)
-            {
-                curPos.x = rc.right - 1;
-                isClamped = true;
-            }
-        }
-        else
-        {
-            curPos.x = rc.left;
-            isClamped = true;
-        }
-        if (curPos.y >= rc.top)
-        {
-            if (curPos.y >= rc.bottom)
-            {
-                curPos.y = rc.bottom - 1;
-                isClamped = true;
-            }
-        }
-        else
-        {
-            curPos.y = rc.top;
-            isClamped = true;
-        }
+		GetCursorPos(&curPos);
+		GetWindowRect(Game::g_wv->hWnd, &rc);
+		bool isClamped = false;
+		if (curPos.x >= rc.left)
+		{
+			if (curPos.x >= rc.right)
+			{
+				curPos.x = rc.right - 1;
+				isClamped = true;
+			}
+		}
+		else
+		{
+			curPos.x = rc.left;
+			isClamped = true;
+		}
+		if (curPos.y >= rc.top)
+		{
+			if (curPos.y >= rc.bottom)
+			{
+				curPos.y = rc.bottom - 1;
+				isClamped = true;
+			}
+		}
+		else
+		{
+			curPos.y = rc.top;
+			isClamped = true;
+		}
 
-        if (isClamped)
-        {
-            SetCursorPos(curPos.x, curPos.y);
-        }
+		if (isClamped)
+		{
+			SetCursorPos(curPos.x, curPos.y);
+		}
 	}
 
-    void IN_RawMouseMove()
-    {
-        static Game::dvar_t* r_fullscreen = Game::Dvar_FindVar("r_fullscreen");
+	int RawMouse::mouseRawX = 0;
+	int RawMouse::mouseRawY = 0;
 
-        if (GetForegroundWindow() == Game::g_wv->hWnd)
-        {
-            if (r_fullscreen->current.enabled)
-                IN_ClampMouseMove();
+	BOOL RawMouse::OnRawInput(LPARAM lParam, WPARAM)
+	{
+		UINT dwSize = sizeof(RAWINPUT);
+		static BYTE lpb[sizeof(RAWINPUT)];
 
+		GetRawInputData((HRAWINPUT)lParam, RID_INPUT, lpb, &dwSize, sizeof(RAWINPUTHEADER));
 
-        }
-    }
+		auto* raw = reinterpret_cast<RAWINPUT*>(lpb);
+		if (raw->header.dwType == RIM_TYPEMOUSE)
+		{
+			// Is there's really absolute mouse on earth?
+			if (raw->data.mouse.usFlags & MOUSE_MOVE_ABSOLUTE)
+			{
+				mouseRawX = raw->data.mouse.lLastX;
+				mouseRawY = raw->data.mouse.lLastY;
+			}
+			else
+			{
+				mouseRawX += raw->data.mouse.lLastX;
+				mouseRawY += raw->data.mouse.lLastY;
+			}
+		}
 
-    Dvar::Var Mouse_RawInput;
+		return TRUE;
+	}
 
-    void IN_MouseMove()
-    {
-        if (Mouse_RawInput.get<bool>())
-        {
-            IN_RawMouseMove();
-        }
-        else
-        {
-            Game::IN_MouseMove();
-        }
-    }
+	void RawMouse::IN_RawMouseMove()
+	{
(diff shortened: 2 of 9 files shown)


src/Components/Loader.hpp
```diff
@@ -137,3 +137,5 @@ namespace Components
 
 #include "Modules/Gamepad.hpp"
 #include "Modules/ScriptExtension.hpp"
+
```
diamante0018, line 140: Remove space

src/Components/Modules/RawMouse.cpp
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		if (isClamped)
+		{
+			SetCursorPos(curPos.x, curPos.y);
+		}
+	}
+
+	int RawMouse::mouseRawX = 0;
+	int RawMouse::mouseRawY = 0;
```
diamante0018, line 49: Static class definitions have been traditionally been declared at the top of the cpp file (near dvar declaration)

src/Components/Modules/RawMouse.hpp
```diff
@@ -0,0 +1,20 @@
...
+	class RawMouse : public Component
+	{
+	public:
+		RawMouse();
+	private:
+		static Dvar::Var m_rawinput;
+		static int mouseRawX, mouseRawY;
+
+		static void IN_ClampMouseMove();
+		static BOOL OnRawInput(LPARAM lParam, WPARAM);
```
diamante0018, line 14: We don't use c macros in the project. use regular bool if possible.

src/Game/Functions.hpp
```diff
@@ -1005,6 +1005,21 @@ namespace Game
 	typedef EffectiveStance(__cdecl * PM_GetEffectiveStance_t)(const playerState_s* ps);
 	extern PM_GetEffectiveStance_t PM_GetEffectiveStance;
 
+	typedef int(__cdecl * CL_MouseEvent_t)(int x, int y, int dx, int dy);
+	extern CL_MouseEvent_t CL_MouseEvent;
+
+	typedef void(*IN_RecenterMouse_t)();
```
diamante0018, line 1011: __cdecl *. I know it's redundant but I'd like to keep consistency.

src/Components/Modules/Window.hpp
```diff
@@ -5,6 +5,9 @@ namespace Components
 	class Window : public Component
 	{
 	public:
+		typedef BOOL(WndProcCallback)(WPARAM wParam, LPARAM lParam);
```
diamante0018, line 8: why does the callback need to return a BOOL which is never used?

src/Components/Modules/Window.cpp
```diff
@@ -132,15 +147,21 @@ namespace Components
...
 		{
-			Window::ApplyCursor();
-			return TRUE;
+			return WndMessageCallbacks[Msg](lParam, wParam);
 		}
 
 		return Utils::Hook::Call<BOOL(__stdcall)(HWND, UINT, WPARAM, LPARAM)>(0x4731F0)(hWnd, Msg, wParam, lParam);
 	}
 
+	void Window::EnableDpiAwareness()
```
diamante0018, line 158: Should add a check to see if you should call the actual function I think. Could also check the active pull request on the launcher that aims to fix a dpi related issue.
[launcher](https://github.com/XLabsProject/launcher/blob/develop/src/launcher/main.cpp#L45)

diamante0018 (changes requested):
We are trying to use auto across new project's components and you should too for most local variable declarations like auto isclamped = true; etc...

src/Components/Modules/RawMouse.cpp
```diff
@@ -0,0 +1,165 @@
...
+		}
+		else
+		{
+			Game::IN_MouseMove();
+		}
+	}
+
+	RawMouse::RawMouse()
+	{
+		Utils::Hook(0x475E65, RawMouse::IN_MouseMove, HOOK_JUMP).install()->quick();
```
diamante0018, line 152: Noticed this late sorry.
Use Hook::RedirectJump for this kind of hooking (jump). Other hooks are okay

diamante0018 (changes requested):

src/Components/Modules/Window.hpp
GEEKiDoS: See https://github.com/XLabsProject/iw4x-client/pull/230/files/237a89fa710c721dba1ec801040134f7f3967a34#diff-33e2c4cabc153ecbcb147a717b0832c9a2646074838eb011c079a3c3faf00ddaR152

diamante0018: My bad

src/Components/Modules/Window.cpp
GEEKiDoS: The Library::invokePascal already did the check stuff so I think it's just not needed

src/Components/Modules/RawMouse.cpp
GEEKiDoS: Hook::RedirectJump causes crash here, I think there's issue with Hook::RedirectJump?

diamante0018: Oh wow that's terrible. Did not know that could happen. It's weird because it should be ideal for this kind of hook.

GEEKiDoS: Hook::RedirectJump is for short jmp not for long jmp here I think, I'll keep the original code

diamante0018: Yeah I looked at redirectjump code now. It does not overwrite the jump operand opcdode.

src/Components/Modules/RawMouse.cpp
```diff
@@ -2,7 +2,9 @@
 
 namespace Components
 {
-	Dvar::Var RawMouse::m_rawinput;
```
diamante0018: Sorry if I did not specify what is the naming convention.
There is a file in the root of the project directory of iw4x called CODESTYLE
Static class members should begin with an upper case letter.
For Dvars unfortunately it's a bit inconsistent across the project so I apologize if you got confused.
We usually keep the underscore for dvar/game variable/function names.
In this case: `M_RawInput` is correct just as `MRawInput`.

GEEKiDoS: Sorry, my bad. I see now

src/Components/Modules/RawMouse.hpp
```diff
@@ -7,8 +7,8 @@ namespace Components
 	public:
 		RawMouse();
 	private:
-		static Dvar::Var useRawInput;
-		static int mouseRawX, mouseRawY;
+		static Dvar::Var M_RawInput;
+		static int MouseRawX, MouseRawY;
```
GEEKiDoS, line 11: Is it right now?

diamante0018: Lovely

src/Components/Modules/RawMouse.cpp
```diff
@@ -0,0 +1,166 @@
...
+				MouseRawY += raw->data.mouse.lLastY;
+			}
+		}
+
+		return TRUE;
+	}
+
+	void RawMouse::IN_RawMouseMove()
+	{
+		static auto r_fullscreen = Dvar::Var("r_fullscreen");
```
Rackover, line 78: this var can probably be fetched earlier and cached instead of fetched in every rawmousemove!

Rackover (commented):
pretty cool!

src/Components/Modules/RawMouse.cpp
diamante0018: There is the `static` keyword which has a very specific meaning when used inside a function. It's only fetched once.

src/Game/Structs.hpp
```diff
@@ -7473,6 +7473,28 @@ namespace Game
...
+		unsigned int sysMsgTime;
+	};
+
+	struct WinMouseVars_t
+	{
+		int oldButtonState;
+		tagPOINT oldPos;
+		bool mouseActive;
+		bool mouseInitialized;
+	};
```
diamante0018, line 7495: static assert sizeof is 0x10 (16) please

src/Game/Functions.cpp
```diff
@@ -543,6 +550,12 @@ namespace Game
 
 	level_locals_t* level = reinterpret_cast<level_locals_t*>(0x1A831A8);
 
+	WinVars_t* g_wv = reinterpret_cast<WinVars_t*>(0x64A3AC8);
```
diamante0018, line 553: I might be wrong but 0x64A3AC8 has no xrefs for me. Are you sure it's correct? Thanks for checking

diamante0018 (changes requested):
I tested this in-game and I am happy with the changes you propose.
Could you address these last few requested changes I just made?
Then I think we are good to merge!
Thanks

src/Components/Modules/RawMouse.cpp
```diff
@@ -0,0 +1,166 @@
...
+		Utils::Hook(0x475E65, RawMouse::IN_MouseMove, HOOK_JUMP).install()->quick();
+		Utils::Hook(0x475E8D, RawMouse::IN_MouseMove, HOOK_JUMP).install()->quick();
+		Utils::Hook(0x475E9E, RawMouse::IN_MouseMove, HOOK_JUMP).install()->quick();
+
+		Utils::Hook(0x467C03, RawMouse::IN_Init, HOOK_CALL).install()->quick();
+		Utils::Hook(0x64D095, RawMouse::IN_Init, HOOK_JUMP).install()->quick();
+
+		Dvar::OnInit([]()
+		{
+			RawMouse::M_RawInput = Dvar::Register<bool>("m_rawinput", true, Game::dvar_flag::DVAR_ARCHIVE, "Use raw mouse input, use in_restart to take effect if not enabled.");
```
diamante0018, line 160: Could you replace the dvar description with
`Use raw mouse input. Improves accuracy & has better support for higher polling rates.`
We had a discussion in the development channels and this seems to have been the best description in terms of ease of understanding from a user point of view.

src/Game/Functions.cpp
GEEKiDoS: It seems to be right, but seems like the game doesnt use anything other than hWnd and hInstance

GEEKiDoS: I think I will just use Window::GetWindow() instead and remove this struct/g_wv

diamante0018 (approved):

PR #234, opened 2022-05-03
Use master server, fallback to node if not reached

mjkzy:
this will probably spark a discussion, but after talking to Jebus, I'm pretty sure this should happen. this uses `master.xlabs.dev:20810` as the master server already has support for IW4x and if the master server is not reached within 5 seconds, a toast will appear notifying that node is being used and will fallback to using node servers. node doesn't get used on the client until this point is reached. node is then disabled once going back into the server browser again.

im no expert at c++, so please review

### DIFF
--- a/src/Components/Modules/ServerList.cpp
+++ b/src/Components/Modules/ServerList.cpp
@@ -282,7 +282,8 @@ namespace Components
 			Game::netadr_t masterServerAddr;
 			if (ServerList::GetMasterServer(masterServerAddr))
 			{
-				Logger::Print("A valid master server was found at %s:%u\n", masterServerName, masterPort);
+				Toast::Show("cardicon_headshot", "", "Fetching servers...", 3000);
+				useMasterServer = true;
 
 				ServerList::RefreshContainer.awatingList = true;
 				ServerList::RefreshContainer.awaitTime = Game::Sys_Milliseconds();
@@ -291,15 +292,6 @@ namespace Components
 
 				Logger::Print("Sending serverlist request to master\n");
 				Network::SendCommand(ServerList::RefreshContainer.host, "getservers", Utils::String::VA("IW4 %i full empty", PROTOCOL));
-			}
-			else
-			{
-				// this should only be getting called if no master server is found or reached
-				Logger::Print("No valid master server was found, using node as fallback\n");
-
-				useMasterServer = false;
-
-				Node::Synchronize();
 			}
 		}
 		else if (ServerList::IsFavouriteList())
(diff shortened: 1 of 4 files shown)


src/Components/Modules/ServerList.cpp
```diff
@@ -733,6 +738,20 @@ namespace Components
...
 	}
 
+	bool ServerList::GetMasterServer(const char* ip, int port, Game::netadr_t& address)
+	{
+		return Game::NET_StringToAdr(Utils::String::VA("%s:%u", ip, port), &address);
+	}
+
+	bool ServerList::IsServerListOpen()
+	{
+		Game::menuDef_t* menu = Game::Menus_FindByName(Game::uiContext, "pc_join_unranked");
```
diamante0018, line 748: auto* please

src/Components/Modules/ServerList.cpp
```diff
@@ -792,11 +811,9 @@ namespace Components
 		});
 
 		// Set default masterServerName + port and save it 
-#ifdef USE_LEGACY_SERVER_LIST
-		Utils::Hook::Set<char*>(0x60AD92, "127.0.0.1");
+		Utils::Hook::Set<char*>(0x60AD92, "master.xlabs.dev");
```
diamante0018, line 814: const char* please

diamante0018 (changes requested):

src/Components/Modules/ServerList.cpp
mjkzy: oops, just c&p'd the code :P

src/Components/Modules/ServerList.cpp
```diff
@@ -274,22 +269,23 @@ namespace Components
...
-#ifdef USE_LEGACY_SERVER_LIST
-			ServerList::RefreshContainer.awatingList = true;
-			ServerList::RefreshContainer.awaitTime = Game::Sys_Milliseconds();
+			const auto masterPort = Dvar::Var("masterPort").get<int>();
+			const auto masterServerName = Dvar::Var("masterServerName").get<const char*>();
 
-			int masterPort = Dvar::Var("masterPort").get<int>();
-			const char* masterServerName = Dvar::Var("masterServerName").get<const char*>();
+			Game::netadr_t masterServerAddr;
+			if (ServerList::GetMasterServer(masterServerName, masterPort, masterServerAddr))
```
diamante0018, line 276: Responding to your comment, this seems like it's a check to see if the values contained in the two dvars are valid.
I would leave it but flip this if the other way around
if (`!`ServerList::GetMasterServer(masterServerName, masterPort, masterServerAddr))
{
Logger::print("Could not resolve address for %s:%u", ip, port);
}

I think logging failures would be an improvement.

mjkzy: i like this more, will commit soon

diamante0018:
As far as I'm concerned this could be merged as it's a nice feature. Would appreciate if someone else reviewed source.

diamante0018 (approved):

PR #236, opened 2022-05-03
Restore clantags

diamante0018:


diamante0018:
I tested this and added calls to the text renderer component in order to ensure the clan tag is sanitized.
Requires branding pr to be merged first

PR #239, opened 2022-05-03
[Bullet] Add user requested features

diamante0018:


src/Components/Modules/Bullet.cpp
```diff
@@ -0,0 +1,59 @@
...
+	Game::dvar_t* Bullet::BGBulletRange;
+
+	float Bullet::BG_GetSurfacePenetrationDepthStub(const Game::WeaponDef* weapDef, int surfaceType)
+	{
+		assert(weapDef != nullptr);
+		assert(weapDef->penetrateType != Game::PenetrateType::PENETRATE_TYPE_NONE);
+		assert(weapDef->penetrateType < Game::PenetrateType::PENETRATE_TYPE_COUNT);
+		assert(static_cast<size_t>(surfaceType) < Game::materialSurfType_t::SURF_TYPE_COUNT);
+
+		if (const auto penetrationDepth = BGSurfacePenetration.get<float>(); penetrationDepth != 0.0f)
```
Laupetin, line 15: Never `!=` or `==` with floating points :P Use `>` instead

Laupetin, line 15: I don't really like having semicolon separated if statements. I'd move the assignment to the line before.
(With an exception for loops at times depending on the situation)

src/Components/Modules/Bullet.cpp
```diff
@@ -0,0 +1,59 @@
...
+		{
+			// Custom depth
+			return penetrationDepth;
+		}
+
+		// Game's code
+		if (surfaceType != 0)
```
Laupetin, line 22: SURF_TYPE_DEFAULT

Laupetin (commented):
Minor stuff, looks good otherwise

PR #250, opened 2022-05-04
Update gamepad patch dvar flags

Laupetin:
- Make dvars gpad_present, gpad_in_use readonly since they should only be set from cpp code
- Make dvars gpad_stick_pressed_hysteresis, gpad_stick_pressed, gpad_stick_deadzone_max, gpad_stick_deadzone_min, gpad_button_deadzone, gpad_button_lstick_deflect_max, gpad_button_rstick_deflect_max, gpad_use_hold_time saved to config to make it easier for user to configure them


diamante0018 (approved):
10/10

PR #251, opened 2022-05-04
[Logger] Use c++20 & more

diamante0018:


src/Components/Modules/Logger.cpp
```diff
@@ -14,17 +14,30 @@ namespace Components
...
+	{
+		char buffer[4096] = {0};
+
+		va_list va;
+		va_start(va, message);
+		_vsnprintf_s(buffer, _TRUNCATE, message, va);
+		va_end(va);
+
+		const auto customMsg = Utils::String::VA("Debug:\n    %s\nFile:    %s\nLine:    %u\n\n", buffer, filename, line);
+		Logger::MessagePrint(Game::conChannel_t::CON_CHANNEL_DONT_FILTER, {customMsg});
```
Laupetin, line 40: I can see how logging the filename and line might be helpful for some of the debug messages but 5 lines for a single log message are quite excessive. In case of for example the gamepad these would be spammed quite often so i would at least not use this way of logging for all logging statements and offer an alternative that only outputs as a single line.

IIRC your goal for this patch was to offer a way of logging without `#ifdef DEBUG` all the time but this is a bit too much for some places imo

src/Components/Modules/MapDump.cpp
```diff
@@ -78,7 +78,7 @@ namespace Components
 
 		void parseVertices()
 		{
-			Logger::Print("Parsing vertices...\n");
+			LOG_DEBUG("Parsing vertices...");
```
Laupetin, line 81: This gives valid feedback over the progress of the mapdump. I'm not sure why this should be Debug builds only.
As i mentioned earlier, please go over the places you substituted a logging statement for your newly introduced debug log and think about whether the logging call *really* only provides value when debugging the patch.
The goal is not to have no logging anymore when in release mode.

I will not comment this on all occations i disagree with so please check them all.

Laupetin (dismissed):

src/Components/Loader.cpp
```diff
@@ -119,10 +119,10 @@ namespace Components
 		std::reverse(Loader::Components.begin(), Loader::Components.end());
 		for (auto component : Loader::Components)
 		{
-#ifdef DEBUG
+#ifdef _DEBUG
```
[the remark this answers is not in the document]
diamante0018: Sorry I don't know why I did it either

src/Components/Modules/Logger.cpp
diamante0018: Will add new macro

src/Components/Modules/MapDump.cpp
diamante0018: Restored them

diamante0018:
Since we know have the c+20 standard available I re-did the whole thing.

Dss0 (approved):

Dss0 (approved):

PR #254, opened 2022-05-05
Refactor scheduler

diamante0018:


src/Components/Modules/ServerList.cpp
```diff
@@ -911,7 +911,7 @@ namespace Components
 		UIScript::AddOwnerDraw(253, ServerList::UpdateGameType);
 
 		// Add frame callback
-		Scheduler::OnFrame(ServerList::Frame);
+		Scheduler::OnGameInitialized(ServerList::Frame, Scheduler::Pipeline::CLIENT);
```
mjkzy, line 914: wouldn't this be Loop and not OnGameInitialized?

diamante0018: That's not the only mistake I made. Game does not even boot.

momo5502 (approved):

Dss0 (approved):

PR #255, opened 2022-05-05
Make keybind fields in menus work with controller keys

Laupetin:
Before keystrokes in an active bind field could not bind controller keys because the values of the remapped menu keys were taken.
Now all controller buttons can be rebound using these menu fields.
The menu fields show KBM keys whenever KBM is used and gamepad keys whenever gamepad is in use.

Additionally whenever a gamepad button is rebound using a menu field, the button config is set to custom so the gamepad options show an empty value (indicating custom values) for button layout

### DIFF
--- a/src/Components/Modules/Gamepad.cpp
+++ b/src/Components/Modules/Gamepad.cpp
@@ -1033,6 +1033,13 @@ namespace Components
 
 	void Gamepad::UI_GamepadKeyEvent(const int gamePadIndex, const int key, const bool down)
 	{
+		// If we are currently capturing a key for menu bind inputs then do not map keys and pass to game
+		if (*Game::g_waitingForKey)
+		{
+			Game::UI_KeyEvent(gamePadIndex, key, down);
+			return;
+		}
+
 		for (const auto& mapping : controllerMenuKeyMappings)
 		{
 			if (mapping.controllerKey == key)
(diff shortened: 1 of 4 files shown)


src/Components/Modules/Gamepad.cpp
```diff
@@ -1818,6 +1825,35 @@ namespace Components
...
 	}
 
+	void __declspec(naked) Gamepad::Key_GetCommandAssignmentInternal_Stub()
+	{
+		__asm
+		{
+			push eax
+			pushad
+			
+			push [esp + 0x20 + 0x4 + 0x8] // keyNums
```
diamante0018, line 1835: I think it's a bit excessive. 0x24 is pretty self explanatory imo for pushad push eax

src/Components/Modules/Gamepad.cpp
```diff
@@ -1777,19 +1784,19 @@ namespace Components
 		return command;
 	}
 
-	int Gamepad::Key_GetCommandAssignmentInternal_Hk(const char* cmd, int (*keys)[2])
+	int Gamepad::Key_GetCommandAssignmentInternal([[maybe_unused]] int localClientNum, const char* cmd, int (*keys)[2])
```
diamante0018, line 1787: Very good

diamante0018 (approved):

src/Components/Modules/Gamepad.cpp
Laupetin: i guess haha. I had a previous draft of one commit where i had a comment explaining every number so funny you call that excessive already :P

PR #260, opened 2022-05-07
fix visible server list refreshing

mjkzy:
this should fix changing the source (Internet, Local, Favorites) category of the server list and having servers still in the list and visible list not properly working. this reimplements the refreshing on visible list, but it must be specified to refresh. default value for refresh is false as defined in the header.


diamante0018:
Code looks good. Going to test later before merging.

diamante0018 (approved):
This fixed the issue reported in the xlabs discord. Thanks!

PR #262, opened 2022-05-08
Implement kill command

diamante0018:
Should mainly serve as a poc on how to edit tls variables since this is the first time it's done in iw4x source (I think).
Otherwise, it's very cool because this is one command I could not reimplement personally when I created clientcommand component.


diamante0018:
Closing this for now as it's somewhat not that useful

PR #263, opened 2022-05-10
only return node frame on non-dedi instances with master server on

mjkzy:
i read the Dedicated::IsEnabled() check way too fast and realized that SVLanOnly is why it exists.

### DIFF
--- a/src/Components/Modules/Node.cpp
+++ b/src/Components/Modules/Node.cpp
@@ -167,7 +167,7 @@ namespace Components
 
 	void Node::RunFrame()
 	{
-		if (ServerList::useMasterServer) return;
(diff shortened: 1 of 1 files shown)


diamante0018 (changes requested):
You are checking for !Dedicated::IsEnabled() on line 170 and then again on line 171.
Please recheck your logic and detangle these if statements.

mjkzy:
the point is that node **should** be running every frame on the dedi. this is how servers communicate to the node default servers shipped with the client to add themselves onto the node server list. that check is suppose to see if `sv_lanOnly` is enabled, which only applies to dedis and it should be stopping network-related stuff

diamante0018:
I see. Makes sense

diamante0018 (approved):

PR #265, opened 2022-05-10
Add chat callback

diamante0018:
```
init()
{
	OnPlayerSay( ::myCallback );
}
```

```
myCallback( message, mode )
{
	printConsole("Player said: ", message, "\n");
	return false; // false will mute the player, true will allow the message
}
```

Credit goes to [fed](https://github.com/fedddddd)


diamante0018:
Drafting

Dss0 (approved):

PR #332, opened 2022-06-25
[UserInfo] Fix

### LINKED ISSUE #327: setName and resetName functions don't work in S&D
Not sure if this is a known issue but the new setName and resetName gsc functions do not work when the game mode is search and destroy. Everything works fine when it's team deathmatch or free for all though. We have the exact same issue on IW5, the author of both the codes is the same so I suppose there's something wrong in the code.

diamante0018:
Closes #327


diamante0018:
I closed this as a workaround from gsc can be implemented and requires no code changes

PR #353, opened 2022-07-03
[Logger] add new line

diamante0018:


src/Components/Modules/Logger.cpp
```diff
@@ -54,7 +54,8 @@ namespace Components
 		const auto msg = std::vformat(fmt, args);
 		const auto out = std::format("Debug:\n    {}\nFile:    {}\nLine:    {}\n", msg, loc.file_name(), loc.line());
 #else
-		const auto out = "^2" + std::vformat(fmt, args);
+		const auto msg = std::vformat(fmt, args);
+		const auto out = std::format("^2{}\n", msg);
```
Rackover, line 58: ^2  not needed in zonebuilder! any way to remove it when using zonebuilder or stdout? cause it's kind of a pain to have them everywhere in the logs for no reason :v

Dss0 (approved):

PR #362, opened 2022-07-17
Prefix custom GSC functions with "iw4x_"

Rackover:
Now this is up for discussion - but the idea is to move our GSC functions to our own "namespace" to stop breaking half the mods in the planet that had a function called "god" or "ufo", and at the same time make mods a bit more futureproof (more easily identify which mods rely on iw4x and which don't, and not shit in the global GSC namespace in case other mw2 clients might want to use or might have used such named methods too, while not necessarily functioning the same). A bit like vendor prefixes for browsers?

I thought for a time about excluding bot commands from that restriction, not to break compatibility with previous versions of BotWarfare, but to be fair maybe it should be the other way around - and I'll PR botwarfare if this PR gets merged so it gets updated & functional as soon as possible 🙂


Rackover:
This PR includes deprecation warnings for some functions so that modders can have a bit of time to update their mods. In the meantime the mods will remain functionals, but warnings will show.
image
image

In the current state of the PR, BotWarfare 2.1.0 is fully functional - although it prints a few warnings upon compilation.

src/Components/Modules/GSC/Script.cpp
```diff
@@ -4,8 +4,36 @@
...
 {
 	std::string Script::ScriptName;
+	std::string Script::ClientPrefix = "iw4x_";
 	std::unordered_map<std::string, Game::BuiltinFunctionDef> Script::CustomScrFunctions;
 	std::unordered_map<std::string, Game::BuiltinMethodDef> Script::CustomScrMethods;
+
+	// This was added on the 17th of July 2022 to help transition current mods to
+	// the new prefixed functions. Look at the clock! If it's more than three months
+	// later than this date... remove this!
+	std::unordered_set<std::string> Script::DeprecatedFunctionsAndMethods = {
```
diamante0018, line 14: Format

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
 		Game::GScr_LoadGameTypeScript();
 	}
 
-	void Script::AddFunction(const char* name, Game::BuiltinFunction func, int type)
+	void Script::AddFunction(const char* name, Game::BuiltinFunction func, int type, bool addClientPrefix)
 	{
 		Game::BuiltinFunctionDef toAdd;
-		toAdd.actionString = name;
+		std::string prefixedString = Utils::String::ToLower(addClientPrefix ? (Script::ClientPrefix + name).data() : name);
```
diamante0018, line 307: This is ugly to look at, mainly because of the .data() hack to make the ternary operator compile
```
std::string functionName= name;
if ( addClientPrefix )
{
    functionName.insert(0, Script::ClientPrefix)
}

Game::BuiltinFunctionDef toAdd;
...
CustomScrFunctions.insert_or_assign(Utils::String::ToLower(functionName), toAdd);

File: src/Components/Modules/GSC/Script.cpp [hunk 3]
```diff
@@ -273,31 +301,48 @@ namespace Components
...
 		toAdd.actionFunc = func;
 		toAdd.type = type;
 
-		CustomScrFunctions.insert_or_assign(Utils::String::ToLower(name), toAdd);
+		CustomScrFunctions.insert_or_assign(prefixedString, toAdd);
 	}
 
-	void Script::AddMethod(const char* name, Game::BuiltinMethod func, int type)
+	void Script::AddMethod(const char* name, Game::BuiltinMethod func, int type, bool addClientPrefix)
 	{
```
diamante0018, line 316: same as above

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
...
 
-		CustomScrMethods.insert_or_assign(Utils::String::ToLower(name), toAdd);
+		CustomScrMethods.insert_or_assign(prefixedString, toAdd);
+	}
+
+	bool Script::IsDeprecated(const std::string& name) {
+		return Script::DeprecatedFunctionsAndMethods.find(name) != Script::DeprecatedFunctionsAndMethods.end();
 	}
 
 	Game::BuiltinFunction Script::BuiltIn_GetFunctionStub(const char** pName, int* type)
```
diamante0018, line 330: format

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
...
+		toAdd.actionString = nullptr; // We don't need it
 		toAdd.actionFunc = func;
 		toAdd.type = type;
 
-		CustomScrMethods.insert_or_assign(Utils::String::ToLower(name), toAdd);
+		CustomScrMethods.insert_or_assign(prefixedString, toAdd);
+	}
+
+	bool Script::IsDeprecated(const std::string& name) {
+		return Script::DeprecatedFunctionsAndMethods.find(name) != Script::DeprecatedFunctionsAndMethods.end();
```
diamante0018, line 327: https://en.cppreference.com/w/cpp/container/unordered_map/contains

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
...
 	}
 
 	Game::BuiltinFunction Script::BuiltIn_GetFunctionStub(const char** pName, int* type)
 	{
 		if (pName != nullptr)
 		{
-			const auto got = Script::CustomScrFunctions.find(Utils::String::ToLower(*pName));
+			std::string name = Utils::String::ToLower(*pName);
+
+			if (IsDeprecated(name)) {
```
diamante0018, line 336: format

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
...
+				Toast::Show("cardicon_gumby", "WARNING!", Script::ScriptName + " uses the deprecated function " + name, 2048);
```
diamante0018, line 337: std::format()

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +301,48 @@ namespace Components
...
+				Logger::Print(Game::CON_CHANNEL_PARSERSCRIPT, "*** DEPRECATION WARNING ***\n");
+				Logger::PrintError(Game::ERR_SCRIPT, "^1Attempted to execute deprecated builtin {} from {}! This method or function should be prefixed with 'iw4x_'. Automatically fixed the function for now, this will not be the case in future releases! Please update your mod!\n", name, Script::ScriptName);
```
diamante0018, line 339: remove ^1 it's already in printerror

src/Components/Modules/GSC/Script.cpp
```diff
@@ -321,7 +366,18 @@ namespace Components
 	{
 		if (pName != nullptr)
 		{
-			const auto got = Script::CustomScrMethods.find(Utils::String::ToLower(*pName));
+			std::string name = Utils::String::ToLower(*pName);
+
+			if (IsDeprecated(name)) {
```
diamante0018, line 371: same comments as above

src/Components/Modules/UserInfo.cpp
```diff
@@ -59,7 +59,7 @@ namespace Components
 			Game::ClientUserinfoChanged(ent->s.number);
 		});
 
-		Script::AddMethod("ResetName", [](Game::scr_entref_t entref)  // gsc: self ResetName()
+		Script::AddMethod("ResetName", [](Game::scr_entref_t entref)  // gsc: self iw4x_ResetName()
```
diamante0018, line 62: rebase to origin/develop and add setclanname and resetclanname please

diamante0018 (changes requested):

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +302,59 @@ namespace Components
...
-		toAdd.actionString = name;
+		toAdd.actionString = nullptr; // We don't need it
 		toAdd.actionFunc = func;
 		toAdd.type = type;
 
-		CustomScrMethods.insert_or_assign(Utils::String::ToLower(name), toAdd);
+		CustomScrMethods.insert_or_assign(Utils::String::ToLower(functionName), toAdd);
+	}
+
+	bool Script::IsDeprecated(const std::string& name) {
```
diamante0018, line 337: Format

src/Components/Modules/GSC/Script.cpp
```diff
@@ -321,7 +378,19 @@ namespace Components
 	{
 		if (pName != nullptr)
 		{
-			const auto got = Script::CustomScrMethods.find(Utils::String::ToLower(*pName));
+			std::string name = Utils::String::ToLower(*pName);
```
diamante0018, line 381: auto name = please

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +302,59 @@ namespace Components
...
+	bool Script::IsDeprecated(const std::string& name) {
+		return Script::DeprecatedFunctionsAndMethods.contains(name);
 	}
 
 	Game::BuiltinFunction Script::BuiltIn_GetFunctionStub(const char** pName, int* type)
 	{
 		if (pName != nullptr)
 		{
-			const auto got = Script::CustomScrFunctions.find(Utils::String::ToLower(*pName));
+			std::string name = Utils::String::ToLower(*pName);
```
diamante0018, line 345: auto name = please

diamante0018 (changes requested):

src/Components/Modules/GSC/Script.cpp
```diff
@@ -273,31 +302,60 @@ namespace Components
...
 	{
+		std::string functionName = name;
+		if (addClientPrefix)
+		{
+			functionName.insert(0, Script::ClientPrefix);
+		}
+
 		Game::BuiltinMethodDef toAdd;
-		toAdd.actionString = name;
+		toAdd.actionString = nullptr; // We don't need it
```
Laupetin, line 322: Might still be helpful for debugging messages the game outputs maybe? Is there any advantage of setting it to nullptr. You could also set it to the ptr from the std::string that is persisted in the CustomScrMethods map if you want it to have the prefix.

src/Components/Modules/GSC/Script.cpp
```diff
@@ -321,7 +379,19 @@ namespace Components
...
 		if (pName != nullptr)
 		{
-			const auto got = Script::CustomScrMethods.find(Utils::String::ToLower(*pName));
+			auto name = Utils::String::ToLower(*pName);
+
+			if (IsDeprecated(name)) 
+			{
+				Toast::Show("cardicon_gumby", "WARNING!", std::format("{} uses the deprecated method {}", Script::ScriptName, name), 2048);
+				Logger::Print(Game::CON_CHANNEL_PARSERSCRIPT, "*** DEPRECATION WARNING ***\n");
+				Logger::PrintError(Game::ERR_SCRIPT, "Attempted to execute deprecated builtin {} from {}! This function or method should be prefixed with 'iw4x_'. Automatically fixed the function for now, this will not be the case in future releases! Please update your mod!\n", name, Script::ScriptName);
```
Laupetin, line 380: I wouldn't tell the user it's "fixed for now" because they might think it doesn't matter and they don't care. Tell them what they need to do.

src/Components/Modules/GSC/Script.hpp
```diff
@@ -7,17 +7,19 @@ namespace Components
...
-		static void AddMethod(const char* name, Game::BuiltinMethod func, int type = 0);
+		static void AddFunction(const char* name, Game::BuiltinFunction func, int type = 0, bool addClientPrefix = true);
+		static void AddMethod(const char* name, Game::BuiltinMethod func, int type = 0, bool addClientPrefix = true);
 
 		static Game::client_t* GetClient(const Game::gentity_t* gentity);
 
 		static const char* GetCodePosForParam(int index);
 
 	private:
+		static std::string ClientPrefix;
```
Laupetin, line 18: can be constexpr instead of static, it won't change on runtime

src/Components/Modules/GSC/Script.cpp
diamante0018: Unused. We call scr_registerfunction with the name from the map. The struct could be changed and we could use a custom one without actionstring

src/Components/Modules/GSC/Script.hpp
diamante0018: a constexpr std::string? Why? It might make sense if it was a const char*, but in this case, it's useful if it stays a std::string.

Laupetin: constexpr const char* --- nobody said it has to be std::string. Also even if you leave it as std::string you could make it const then

diamante0018: const std::string  sounds good.

Rackover:
Mods will not be able to support both iw4x versions (`if getdvar("version") == "x.x.x" {...} else {...}`) unfortunately due to the way GSC works (needs all function names to be resolved at compile time, no JIT), so they will have to update eventually once the grace period / deprecation period is over 🙂

diamante0018 (approved):

PR #367, opened 2022-07-18
Disable additional zones for vanilla maps

Rackover:
Vanilla maps don't need team FF, so let the arena specify if it's needed or not and if not let's not load additional zones for nothing.
This improves loading time by quite a bit i've heard.

Tested on iw3 and iw4 maps with success, but this still needs a bit more testing


diamante0018 (approved):

Rackover:
Wait for rawfile PR before merging...

PR #373, opened 2022-07-21
Update mongoose

diamante0018:


diamante0018:
Download code should be re written, possibly using lib curl. I'm not doing this anytime soon

PR #393, opened 2022-07-28
Json11 => NlohmannJson

Rackover:
Removes json11 and uses nlohmann json instead (better support for unsigned types and other types)
API is almost the same, had to do a few changes here and there
Zonebuilder custom json material format has been removed since it's unused - no tools produce these materials anymore and ZB/I3X use dot iw4xmaterial binary format... for now!

Tried to build a map and it worked, played on it and it worked, but it could be good to test like Bans and Nodes and so so to make sure nothing has been butchered!


src/STDInclude.hpp
```diff
@@ -95,6 +94,12 @@ using namespace std::literals;
 	#undef min
 #endif
 
+// Needs to be included after the nominmax above ^
+#ifdef snprintf
+	#undef snprintf
+#endif
```
diamante0018, line 100:
```cpp
#pragma warning(push)
#pragma warning(disable: 4100)
#include <json.hpp>
#pragma warning(pop)
``` 
I believe this should do it.

.gitmodules
```diff
@@ -35,3 +31,7 @@
 [submodule "deps/GSL"]
 	path = deps/GSL
 	url = https://github.com/microsoft/GSL.git
+[submodule "deps/nlohmannjson"]
+	path = deps/nlohmannjson
+	url = https://github.com/nlohmann/json.git
+
```
diamante0018, line 37: Please fix this too.

src/STDInclude.hpp
Rackover: It's not a warning it's downright a compilation error

.gitmodules
```diff
@@ -35,3 +31,8 @@
 [submodule "deps/GSL"]
 	path = deps/GSL
 	url = https://github.com/microsoft/GSL.git
+	
+
```
diamante0018, line 35: remove two new lines

src/Components/Modules/AssetInterfaces/IMaterial.cpp
```diff
@@ -332,165 +332,13 @@ namespace Assets
 		header->material = Components::AssetHandler::FindOriginalAsset(this->getType(), name.data()).material;
 	}
 
-	void IMaterial::loadJson(Game::XAssetHeader* header, const std::string& name, Components::ZoneBuilder::Zone* builder)
+	void IMaterial::loadJson(Game::XAssetHeader* header, const std::string& name, Components::ZoneBuilder::Zone* /*builder*/)
```
diamante0018, line 335: `[[maybe_unused]] Components::ZoneBuilder::Zone* builder` please

src/Utils/Json.cpp
```diff
@@ -2,24 +2,28 @@
...
+			return "number_float";
+		case nlohmann::json::value_t::boolean:
+			return "boolean";
+		case nlohmann::json::value_t::string:
+			return "string";
+		case nlohmann::json::value_t::array:
+			return "array";
+		case nlohmann::json::value_t::object:
+			return "object";
 		default:
```
diamante0018, line 25: Use switch case fallthrough for null please
```
case nlohmann::json::value_t::null:
default:
              return "null";
              ```

diamante0018 (changes requested):

diamante0018 (approved):

PR #449, opened 2022-08-22
[Download] Nullsub some quirky gsc functions

diamante0018:


Rackover (approved):
I have not tested it but code looks good. If tested, it's good to go

diamante0018:
Can confirm this works on bot warfare and it causes no script errors (checked with developer dvars and everything)

PR #479, opened 2022-09-07
[General] Small cleanup

diamante0018:
For the bots.cpp change it's about consistency because later in the code memset is used again instead of {0};
For connect protocol it's the same thing; that is more consistency.
And lastly, the generate.bat is a feature request ahah.


Rackover (approved):
Looks good but haven't tested

diamante0018:
I did :)))
Changes purely esthetical

PR #553, opened 2022-11-07
[Threading]: Remove problematic patch

diamante0018:


diamante0018:
Gonna actually close this now as it may cause more issue than it solves.
Will come back to this when I can.

PR #584, opened 2022-11-23
[Cache]: Remove unused code

diamante0018:


src/Utils/Cache.cpp
```diff
@@ -27,9 +27,6 @@ namespace Utils
 
 		if (ValidUrl.empty())
 		{
-			InternetSetCookieA("https://onion.casa", "disclaimer_accepted", "1");
```
momo5502, line 30: Is this whole cache stuff still relevant at all?
I mean those lines are clearly irrelevant, but the rest?

momo5502 (approved):

PR #655, opened 2022-12-17
Fix zonebuilder FindAsset always returning -1 (not found) even when t…

Rackover:
…he asset is actually there & fix loaded sound data not using ZAlloc

### DIFF
--- a/src/Components/Modules/ZoneBuilder.cpp
+++ b/src/Components/Modules/ZoneBuilder.cpp
@@ -46,7 +46,7 @@ namespace Components
 
 			if (!found)
 			{
-				Logger::Print("Asset {} of type {} was loaded, but not written!", name, Game::DB_GetXAssetTypeName(subAsset.type));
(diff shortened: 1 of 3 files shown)


src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1007,11 +1007,11 @@ namespace Components
 		assert(data);
 		auto* sound = Utils::Hook::Get<Game::MssSound*>(0x112AE04);
 		auto length = sound->info.data_len;
-		auto allocatedSpace = Utils::Memory::AllocateArray<char>(length);
+		auto allocatedSpace = Game::Z_Malloc(length);
 		memcpy_s(allocatedSpace, length, data, length);
 
 		data = allocatedSpace;
-		sound->data = allocatedSpace;
+		sound->data = reinterpret_cast<char*>(allocatedSpace);
```
diamante0018, line 1014: static_cast<char*>

diamante0018 (changes requested):

diamante0018 (approved):

PR #662, opened 2022-12-21
Zonebuilder directory for support zones

Rackover:
This PR allows zonebuilder to load from zone/zonebuilder (but not the standard client). This way we can have zones there that while very useful for zone builder, are not strictly necessary for the game itself to run!

See https://github.com/XLabsProject/iw4x-rawfiles/pull/37


src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1015,6 +1015,18 @@ namespace Components
 		sound->info.data_ptr = allocatedSpace;
 	}
 
+	HANDLE ZoneBuilder::Sys_CreateFile_FromZonebuilderFolder(const char* dir, const char* zoneFileName)
+	{
+		auto handle = Utils::Hook::Call<HANDLE(const char*, const char*)>(0x4B2EF0)(dir, zoneFileName);
+
+		if (handle == INVALID_HANDLE_VALUE)
+		{
+			handle = Utils::Hook::Call<HANDLE(const char*, const char*)>(0x4B2EF0)("zone\\zonebuilder\\", zoneFileName);
```
diamante0018, line 1024: Limited review capability as I'm from phone. Can you define this function in the System.hpp file?

diamante0018 (changes requested):

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1029,6 +1041,8 @@ namespace Components
 			// Prevent loading textures (preserves loaddef)
 			//Utils::Hook::Set<BYTE>(Game::Load_Texture, 0xC3);
 
+			Utils::Hook(0x5BC832, Sys_CreateFile_FromZonebuilderFolder, HOOK_CALL).install()->quick();
```
diamante0018, line 1044: A detail I noticed now:
I would prefer you stuck to stricly using laupetin's naming style.
`Sys_CreateFile_Stub` please add the explanation to what this function does in a comment.
I'd prefer that other having the summary in the function name.
A comment can have more info, if necessary, like `if the fastfiles can't be found we look in the zonebuilder folder`

diamante0018 (changes requested):

src/Game/System.hpp
```diff
@@ -61,10 +61,13 @@ namespace Game
 
 	typedef void(*Sys_SuspendOtherThreads_t)();
 	extern Sys_SuspendOtherThreads_t Sys_SuspendOtherThreads;
-
+
```
diamante0018, line 64: ? lol ?

diamante0018 (approved):

diamante0018 (approved):

diamante0018 (approved):

PR #674, opened 2022-12-27
[Exception]: Fix

diamante0018:


Rackover (approved):
Have not tested it myself, but @diamante0018  tested it and said it works

PR #748, opened 2023-02-02
[GamePad]: Add one check

diamante0018:
This patch can be later improved on at some later date. But I see this as a pretty effective prevention method for some game mechanics game pad users can potentially gain advantage from.


Rackover (dismissed):
(cf discord)

edit: I disagree with the way this problem is handled because it can introduce a situation where the user's mouse will lock up and they won't know why (if they have a controller plugged in on their system). The cause of that lock up will not be immediatly obvious for the user because we don't have a warning of any kind for when that happens.

I made a twofold suggestion:
- To move the detection threshold from 0.0F to 0.4F here
https://github.com/XLabsProject/iw4x-client/blob/353290aa8809cec56ccffec71f5e8fb495a7b19f/src/Components/Modules/Gamepad.cpp#L1047
This way you can't have "mouse and gamepad at the same time" and if you try to tilt your gamepad _while playing with the mouse_ the tilt will be hard enough to make it unplayable.
- And to disable Aim assist target detection instead of locking up the mouse. This means whenever devices gets switched it's imperceptible to the end user & the main exploit we're targeting is addressed.

I am against the way this PR handles it but did not rally anyone around this suggestion so I'm leaving this comment here for future reference.


tl;dr:  bad idea imo

diamante0018:
> (cf discord)

I disagree

diamante0018:
Oops, didn't mean to re-request a review.

momo5502 (approved):

PR #750, opened 2023-02-04
Zonebuilder update for iw5xport compat

Rackover:
There's a loooot of stuff going on in here. Its not entirely clean yet, some stuff in clipmap uses the wrong naming convention. But this is the code that is necessary to build maps exported from iw5xport basically. I need to export iw3xport too before we can merge this.


diamante0018 (approved):

Rackover:
WHY WAS THIS MERGED 
IT IS NOT READY

PR #772, opened 2023-02-14
IW4OF integration

Rackover:
don't merge it this time!


Rackover:
(yes the branch is fucked, i'm gonna rebase)

.gitmodules
```diff
@@ -35,3 +35,6 @@
 	path = deps/json
 	url = https://github.com/nlohmann/json.git
 	branch = v3.11.2
+[submodule "deps/iw4-open-formats"]
+	path = deps/iw4-open-formats
+	url = https://github.com/Rackover/iw4-open-formats
```
diamante0018, line 40: missing .git extension

src/Components/Modules/ZoneBuilder.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "api.hpp"
```
diamante0018, line 3: move this to .cpp file, the whole project doesn't need to see your api.

deps/premake/iw4-open-formats.lua
```diff
@@ -0,0 +1,73 @@
+iw4of = {
```
diamante0018, line 1: this needs to be re-written almost entirely.
might do it once this pr has no more conflitcs

src/Components/Modules/ZoneBuilder.hpp
Rackover: but that would mean effectively including the HPP in every single asset interface, is that okay ?

Rackover:
ready to merge as soon as premake is good, this will break iw3xport however until iw3xport is updated

src/Components/Modules/ZoneBuilder.hpp
diamante0018: gotcha this can stay for now, as a heads up when making this type of include <> should be preferred because the compiler uses a different search algorithm for the header.

diamante0018 (approved):

PR #775, opened 2023-02-15
IW4Of Fixes

Rackover:
Some stuff we forgot or messed up

### DIFF
--- a/src/Components/Modules/Maps.cpp
+++ b/src/Components/Modules/Maps.cpp
@@ -666,12 +666,24 @@ namespace Components
 		return Utils::Hook::Call<bool(Game::gentity_s*)>(0x5050C0)(ent);
 	}
 
-	int16 Maps::CM_TriggerModelBounds(int modelPointer, Game::Bounds* bounds) {
-#ifdef DEBUG
+	int16 Maps::CM_TriggerModelBounds(unsigned int triggerIndex, Game::Bounds* bounds) {
+
 		Game::MapEnts* ents = *reinterpret_cast<Game::MapEnts**>(0x1AA651C);  // Use me for debugging
-		(void)ents;
-#endif
-		return Utils::Hook::Call<int16(int, Game::Bounds*)>(0x4416C0)(modelPointer, bounds);
+
+		if (ents)
+		{
+			if (triggerIndex >= ents->trigger.count)
+			{
+				Components::Logger::Error(Game::errorParm_t::ERR_DROP, "Invalid trigger index {} in entities exceeds the maximum trigger count ({}) defined in the clipmap. Check your map ents, or your clipmap.", triggerIndex, ents->trigger.count);
+				return 0;
+			}
+			else 
+			{
+				return Utils::Hook::Call<int16(int, Game::Bounds*)>(0x4416C0)(triggerIndex, bounds);
+			}
+		}
+
+		return 0;
 	}
 	
 	Maps::Maps()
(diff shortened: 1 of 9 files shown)


Rackover:
Still not building valid zones btw so wait a bit before merging, it's not over

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -795,9 +795,16 @@ namespace Components
 			}
 		};
 
-		params.work_directory = (*Game::fs_basepath)->current.string;
+		if (Game::fs_basepath && Game::fs_gameDirVar)
```
diamante0018, line 798: this is wrong it should be *fs_basepath && *fs_gameDirVar
the type of this variable is ** (double pointer) so this if statement is never false

diamante0018 (changes requested):

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -535,10 +535,11 @@ namespace Components
...
 	void ZoneBuilder::Zone::addBranding()
 	{
-		constexpr auto* data = "Built using the IW4x Zone:B:uilder Version 4";
-		auto dataLen = std::strlen(data); // + 1 is added by the save code
+		const auto now = std::chrono::system_clock::now();
+		std::string data = std::format("Built using the IW4x ZoneBuilder! {:%d-%m-%Y %H:%M:%OS}", now);
+		auto dataLen = data.size(); // + 1 is added by the save code
 
-		this->branding = {this->zoneName.data(), 0, static_cast<int>(dataLen), data};
+		this->branding = {this->zoneName.data(), 0, static_cast<int>(dataLen), data.data()};
```
diamante0018, line 542: pretty sure UAF for data, have you tested this?

Rackover: Not yet this is a WIP, i have bigger problem right now with remappedtechinquesets not being saved properly, I'll come back to this before this PR is ready :)

diamante0018: Okay, I'll mark this as "request changes" if I can so it does not go unforgotten. this becomes a nulltpr after the function exits.

src/Components/Modules/Maps.cpp
```diff
@@ -666,12 +666,24 @@ namespace Components
 		return Utils::Hook::Call<bool(Game::gentity_s*)>(0x5050C0)(ent);
 	}
 
-	int16 Maps::CM_TriggerModelBounds(int modelPointer, Game::Bounds* bounds) {
-#ifdef DEBUG
+	int16 Maps::CM_TriggerModelBounds(unsigned int triggerIndex, Game::Bounds* bounds) {
```
diamante0018, line 669: Please fix the braces and rename function to contain the word Stub or Hk or Hook

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -535,10 +535,11 @@ namespace Components
 	// Add branding asset
 	void ZoneBuilder::Zone::addBranding()
 	{
-		constexpr auto* data = "Built using the IW4x Zone:B:uilder Version 4";
-		auto dataLen = std::strlen(data); // + 1 is added by the save code
+		const auto now = std::chrono::system_clock::now();
+		std::string data = std::format("Built using the IW4x ZoneBuilder! {:%d-%m-%Y %H:%M:%OS}", now);
```
diamante0018, line 539: TBH I didn't know this was possible but I tested it and it works.
Example output: `Built using the IW4x ZoneBuilder! 16-02-2023 18:51:33`
Just use auto please.

Rackover: Ok so no need to fix UAF or anything ?

diamante0018: What? This is unrelated

diamante0018: UAF is still present a couple of lines after this

Rackover: Did not know what you were talking about with "i tested it and it works"

diamante0018: My bad. What I meant to say is that I didn't know you could format the time like that. But yeah the UAF is present.
🅱️

Rackover:
This is ready as far as i'm concerned, builds maps properly

diamante0018 (dismissed):

diamante0018 (approved):

PR #781, opened 2023-02-17
Revert diamantes last minute modifications

Rackover:
maybe if we did reviews the normal way instead of directly writing onto each other's branches....

### DIFF
--- a/src/Components/Modules/ZoneBuilder.cpp
+++ b/src/Components/Modules/ZoneBuilder.cpp
@@ -535,13 +535,12 @@ namespace Components
 	// Add branding asset
 	void ZoneBuilder::Zone::addBranding()
 	{
-		static std::string zoneBranding;
-
 		const auto now = std::chrono::system_clock::now();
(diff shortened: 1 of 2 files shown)


src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -535,13 +535,12 @@ namespace Components
 	// Add branding asset
 	void ZoneBuilder::Zone::addBranding()
 	{
-		static std::string zoneBranding;
-
 		const auto now = std::chrono::system_clock::now();
-		zoneBranding = std::format("Built using the IW4x ZoneBuilder! {:%d-%m-%Y %H:%M:%OS}", now);
+		
+		std::string zoneBranding = std::format("Built using the IW4x ZoneBuilder! {:%d-%m-%Y %H:%M:%OS}", now);
```
diamante0018, line 540: Use auto like I asked in the other review please.

diamante0018 (changes requested):

diamante0018 (approved):

PR #786, opened 2023-02-20
[Memory/ZB]: Use faster memory allocator

diamante0018:
IW4x makes extensive use of malloc more than any other of our clients combined, it is used so often in almost every function of the ZoneBuilder code & related dependency.
This extra library I am adding is claimed by Microsoft to be very fast & properly benchmarked.
Overall this should be a nice addition that speeds up the runtime speed of iw4x in zone builder mode & regular iw4x as well since we make a lot of allocations and I mean a lot, like one allocation for each single localized string entry, CSV table loaded from raw and many many more.

In DEBUG mode the library will debug the HEAP properly, there is also a secure mode but it's turned off by default and I don't think we need it right now.
See the Microsoft repository for more information.

Because IW4x only uses malloc/calloc in one file (Memory.cpp) this is a painless insertion into our repository.


momo5502:
I'm not a friend of "performance optimizations" that aren't proven through proper profiling.

"we make a lot of allocations and I mean a lot" - what is a lot?
who defines that? are a million allocations a lot? does that even matter? is maybe size more relevant? cause if the same storage is released and requested again, it might even be practically free because the allocator reuses the same slot.

I personally don't like including an additional 3rd party library just for the intagible performance boost it may potentially bring.

However, as I'm not the maintainer of IW4x, I won't reject this PR, but neither am I going to approve it.

diamante0018:
Fair enough. I might as well completely agree unless I do the proper profiling and come back with concrete results.

diamante0018:
Because I am a noob I do not know how to use Valgrind but I did my best to profile using the classic timer technique (taking the current time on the start of a task and subtracting the current time after the task ends yield elapsed time).

Here are my results:
Quite interestingly I was able to profile a noticeable speed improvement only when building Zones with ZoneBuilder.
When profiling DllMain on IW4x to see the time it took to initialize the entirety of IW4x and initialize it I did *not* notice any difference.
As I noted before, one of the most allocation-intensive assets is the localized strings asset, because there is a delete/new operation for each localized asset that is parsed
Current IW4x does complete parsing in around 1500/1600 msec when testing with the custom `IW4x` localized string file from the raw files.
With mimalloc that goes to 1100/1200 msec.
It's about a 20% speed increase.

My approximate conclusions are that using mimalloc provides a speed boost on tasks that require a `rapid` and `high` number of allocations/deallocations. Like allocating space or deleting space for each line of a huge file whilst parsing it.
There is no noticeable speed boost on single large allocations.

diamante0018:
After further consideration it has come to this:
Basically this would only benefit zonebuilder.
Unless we dynamically override malloc/free family using an external DLL by modifying the iw4x.exe import table, the game would still be using the old CRT version of memory functions.
Unless we make an effort to dynamically override iw4x.exe and statically iw4x.dll the gained advantages are limitied to zonebuilder.

PR #791, opened 2023-02-21
Fix snd/sndcurve not using original assets as fallback, bump IW4Of

Rackover:
…and a few other things

### DIFF
--- a/src/Components/Modules/AssetInterfaces/IMaterialTechniqueSet.cpp
+++ b/src/Components/Modules/AssetInterfaces/IMaterialTechniqueSet.cpp
@@ -14,7 +14,7 @@ namespace Assets
 
 		if (!header->data)
 		{
-			assert(false);
+			AssertUnreachable();
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


src/Components/Modules/AssetInterfaces/IMaterialTechniqueSet.cpp
```diff
@@ -12,7 +12,10 @@ namespace Assets
 		if (!header->data) this->loadFromDisk(header, name, builder); // Check if we need to import a new one into the game
 		if (!header->data) this->loadNative(header, name, builder); // Check if there is a native one
 
-		assert(header->data);
+		if (!header->data)
+		{
+			assert(false);
```
diamante0018, line 17: https://github.com/XLabsProject/iw4x-client/blob/develop/src/STDInclude.hpp#L90

diamante0018 (changes requested):

diamante0018 (approved):

PR #799, opened 2023-02-27
Small thing

Rackover:


src/Components/Modules/Renderer.cpp
```diff
@@ -178,29 +179,32 @@ namespace Components
...
+				}
+				else
+				{
+					color[0] = 0.f;
+				}
 
-			std::stringstream str;
-			str << std::format("{}/{:#X} => ", i, i) << (source->input.codeImages[i] == nullptr ? "---" : source->input.codeImages[i]->name) << " " << std::to_string(source->input.codeImageSamplerStates[i]);
-			Game::R_AddCmdDrawText(str.str().data(), std::numeric_limits<int>::max(), font, 15.0f, (height * scale + 1) * (i + 1) + 14.0f, scale, scale, 0.0f, color, Game::ITEM_TEXTSTYLE_NORMAL);
+				std::stringstream str;
```
diamante0018, line 204: I don't know about this one to be honest.

Rackover: It's already there actually

src/Components/Modules/Renderer.cpp
```diff
@@ -178,29 +179,32 @@ namespace Components
 		return Utils::Hook::Call<int(int, float, float, const char*, Game::vec4_t*, int)>(0x005033E0)(a1, a2, a3, Utils::String::VA("%s (^3%s^7)", mat->info.name, mat->techniqueSet->name), color, a6);
 	}
 
-	void ListSamplers()
+	void Renderer::ListSamplers()
 	{
-		static auto* source = reinterpret_cast<Game::GfxCmdBufSourceState*>(0x6CAF080);
+		if (r_listSamplers.get<bool>())
```
diamante0018, line 184: Should return here.

src/Components/Modules/Renderer.cpp
```diff
@@ -178,29 +179,32 @@ namespace Components
...
+			static auto* source = reinterpret_cast<Game::GfxCmdBufSourceState*>(0x6CAF080);
 
-		Game::Font_s* font = Game::R_RegisterFont("fonts/smallFont", 0);
-		auto height = Game::R_TextHeight(font);
-		auto scale = 1.0f;
-		float color[4] = {0.0f, 1.0f, 0.0f, 1.0f};
+			Game::Font_s* font = Game::R_RegisterFont("fonts/smallFont", 0);
+			auto height = Game::R_TextHeight(font);
+			auto scale = 1.0f;
+			float color[4] = { 0.0f, 1.0f, 0.0f, 1.0f };
```
diamante0018, line 191: Unnecessary code change by adding space after { }

src/Components/Modules/Renderer.cpp
diamante0018, line 204: Could you try to do without?

diamante0018 (changes requested):

src/Components/Modules/Renderer.cpp
Rackover: without the stringstream you mean? you sure it would be faster without?

diamante0018: We never use it for string formatting, it should be possible to simply use one std::format.
...Or not?

Rackover: i once again do not understand what you're talking about. Do you not want to use it because "we never use it" or do you have an actual reason why stringstream would be bad in the current situation ?

Rackover: performance wise concatenating strings would be way worse, what are your alternatives & why are you looking for one .?

diamante0018: Why are you brining up string concatenation?
Since you have std::format available can't you use one format string and pass the different things as arguments of std::format. That is what I intended the first time.

diamante0018 (approved):

PR #814, opened 2023-03-06
IW5 DLC Prep

Rackover:
Preparation for the IW5 DLC update.
You may comment but you may not touch !


src/Components/Modules/FastFiles.cpp
```diff
@@ -213,6 +217,23 @@ namespace Components
 		data.push_back(info);
 
 		Game::DB_LoadXAssets(data.data(), data.size(), sync);
+
+		Scheduler::Once([]()
```
diamante0018, line 221: https://github.com/XLabsProject/iw4x-client/blob/develop/src/Components/Modules/Scheduler.cpp#L135

Rackover: i think this is not what i want

diamante0018: I thought otherwise, it seems to be almost identical to your function. Could you please try to use it and see if it works?

src/Utils/Thread.cpp
```diff
@@ -78,45 +78,4 @@ namespace Utils::Thread
 
 		return ids;
 	}
-
```
diamante0018, line 81: I do not approve of the removal of this for no reason whatsoever.

src/Components/Modules/Zones.cpp
```diff
@@ -3479,71 +3479,74 @@ namespace Components
 	{
 		Zones::ZoneVersion = 0;
 
-		Command::Add("decryptImages", [](Command::Params*)
+		if (ZoneBuilder::IsEnabled())
```
diamante0018, line 3482: Fix the formatting of this file.

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1431,8 +1432,10 @@ namespace Components
 				auto list = Utils::IO::ListFiles("zone/techsets", false);
 				int i = 0;
 				int subCount = 0;
-				for (auto it : list)
+				for (auto entry : list)
```
diamante0018, line 1435: use reference, `auto& entry`

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1336,8 +1336,9 @@ namespace Components
 
 				const auto dir = std::format("zone/{}", Game::Win_GetLanguage());
 				auto fileList = Utils::IO::ListFiles(dir, false);
-				for (auto zone : fileList)
+				for (auto entry : fileList)
```
diamante0018, line 1339: use reference, `auto& entry`

src/Components/Modules/UIFeeder.cpp
```diff
@@ -387,8 +394,8 @@ namespace Components
 
 		Events::OnDvarInit([]
 		{
-			UIMapLong = Dvar::Register<const char*>("ui_map_long", "Afghan", Game::DVAR_NONE, "");
-			UIMapName = Dvar::Register<const char*>("ui_map_name", "mp_afghan", Game::DVAR_NONE, "");
```
diamante0018, line 391: Why did you remove the default vaue?

src/Components/Modules/UIFeeder.cpp
```diff
@@ -279,74 +279,81 @@ namespace Components
...
 
 	void UIFeeder::ApplyMap([[maybe_unused]] const UIScript::Token& token, [[maybe_unused]] const Game::uiInfo_s* info)
 	{
 		const auto mapname = Dvar::Var("ui_map_name").get<std::string>();
 
-		Dvar::Var("ui_mapname").set(mapname);
-		Utils::Hook::Call<void(const char*)>(0x503B50)(mapname.data()); // Party_SetDisplayMapName
+		if (!mapname.empty())
+		{
+			Dvar::Var("ui_mapname").set(mapname);
```
diamante0018, line 348: dvar already defined here.
https://github.com/XLabsProject/iw4x-client/blob/develop/src/Game/Dvars.cpp#L86

src/Components/Modules/UIFeeder.cpp
```diff
@@ -279,74 +279,81 @@ namespace Components
...
+					mapName = arena.at("map");
+				}
+
+				if (arena.contains("description"))
+				{
+					description = arena.at("description");
+				}
+			}
+
+			UIMapName.set(Game::SEH_StringEd_GetString(mapName.data()));
```
diamante0018, line 336: Why are you using this instead of https://github.com/XLabsProject/iw4x-client/blob/develop/src/Components/Modules/Localization.hpp#L12

src/Components/Modules/UIFeeder.cpp
```diff
@@ -279,74 +279,81 @@ namespace Components
...
-		Game::UI_UpdateArenas();
-		Game::UI_SortArenas();
+		const auto& maps = Maps::GetCustomMaps();
+		if (index < maps.size())
+		{
+			return maps.at(index).data();
+		}
 
-		if (index < static_cast<unsigned int>(*Game::arenaCount))
+#if DEBUG
```
diamante0018, line 293: Wrong! Use `#ifdef` for checking if a macro exists.

src/Components/Modules/Renderer.cpp
```diff
@@ -17,6 +17,7 @@ namespace Components
 	Dvar::Var Renderer::r_playerDrawDebugDistance;
 	Dvar::Var Renderer::r_forceTechnique;
 	Dvar::Var Renderer::r_listSamplers;
+	Dvar::Var Renderer::r_drawLights;
```
diamante0018, line 20: Why is this changed mixed with an IW5 DLC Prep PR? Put it in another PR please.

src/Components/Modules/ArenaLength.cpp
```diff
@@ -57,7 +60,19 @@ namespace Components
 
 	ArenaLength::ArenaLength()
 	{
-		// Reallocate array
+		// Reallocate  ui_arenaInfos
+		Utils::Hook::Set<char**>(0x4A95F0 + 3, &ArenaLength::NewArenaInfos[0]);
```
diamante0018, line 64: Simplify with just writing `ArenaLength::NewArenaInfos` instead of & and [0]

src/Components/Modules/Maps.cpp
```diff
@@ -11,6 +11,8 @@ namespace Components
 	std::string Maps::CurrentMainZone;
 	std::vector<std::pair<std::string, std::string>> Maps::DependencyList;
 	std::vector<std::string> Maps::CurrentDependencies;
+	std::vector<std::string> Maps::FoundCustomMaps;
+	std::mutex Maps::CustomMapsScanningMutex;
```
diamante0018, line 15: For what reason is there a mutex? Does seem like there is one? Where is the multithreading in this code?

diamante0018 (changes requested):

src/Components/Modules/UIFeeder.cpp
Rackover: Cause Afghan is not a usermap . The whole purpose of this feeder has changed.
The default value is not necessary anymore anyway

src/Components/Modules/Renderer.cpp
Rackover: It's not the only cleanup thing here. There's also the techset dump thing that was superseeded by iw5of, among other things. 
I saw you change some completely unrelated menufiles on the IW5 menu PR to remove a checkClass call, oddly enough it did not strike you as out-of-place enough to warrant a new PR. I'm sure you'll be able to mentally accommodate that this change will cause no harm in this PR, just like the dumptechset one or the checkclass one.

src/Components/Modules/Maps.cpp
diamante0018: Yes.

src/Utils/Thread.cpp
diamante0018: Nope, it's you who has to elaborate as to why you want to remove this, not me.
The code can stay.

Rackover: It's deadcode

src/Components/Modules/UIFeeder.cpp
Rackover: It's older code, did not know about Localization

src/Components/Modules/UIFeeder.cpp
Rackover: i know it's used that way throughout the iw4x codebase so i'll modify it to match, but still this sounds really dumb to me. `DEBUG false` should not trigger the debug blocks. We should use #if instead of #ifdef for these throughout the codebase...

src/Components/Modules/FastFiles.cpp
```diff
@@ -135,6 +137,8 @@ namespace Components
 	// This has to be called only once, when the game starts
 	void FastFiles::LoadInitialZones(Game::XZoneInfo *zoneInfo, unsigned int zoneCount, int sync)
 	{
+		g_loadingInitialZones = Dvar::Register("g_loadingInitialZones", true, 0, "Is still loading initial zones");
```
diamante0018, line 140: `Dvar::Register<bool>` to respect the template argument defined here: https://github.com/XLabsProject/iw4x-client/blob/develop/src/Components/Modules/Dvar.hpp#L37

src/Components/Modules/Maps.cpp
```diff
@@ -578,6 +600,35 @@ namespace Components
...
+					FoundCustomMaps.push_back(zoneName);
+					Components::Logger::Print("Discovered custom map {}\n", zoneName);
+				}
+			}
+		}
+	}
+
+	std::string Maps::GetArenaPath(const std::string& mapName)
+	{
+		return Utils::String::VA("usermaps/%s/%s.arena", mapName.data(), mapName.data());
```
diamante0018, line 629: since the return type is std::string std::format should be used

src/Components/Modules/Maps.cpp
```diff
@@ -578,6 +600,35 @@ namespace Components
...
+		for (const auto& entry : entries)
+		{
+			if (entry.is_directory())
+			{
+				auto zoneName = entry.path().filename().string();
+				auto mapPath = std::format("{}\\{}.ff", entry.path().string(), zoneName);
+				if (Utils::IO::FileExists(mapPath))
+				{
+					FoundCustomMaps.push_back(zoneName);
+					Components::Logger::Print("Discovered custom map {}\n", zoneName);
```
diamante0018, line 621: Remove Components:: as this is already inside said the namespace

src/Components/Modules/Maps.cpp
```diff
@@ -106,7 +107,7 @@ namespace Components
 		if (Maps::UserMap.isValid())
 		{
 			const std::string mapname = Maps::UserMap.getName();
-			const auto* arena = Utils::String::VA("usermaps/%s/%s.arena", mapname.data(), mapname.data());
+			std::string arena = GetArenaPath(mapname);
```
diamante0018, line 110: const auto is preferred

src/Components/Modules/Zones.cpp
```diff
@@ -2222,10 +2222,10 @@ namespace Components
 			push 4189AEh
 			retn
 
-		returnSafe:
```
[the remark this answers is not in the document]
Rackover: i formatted the whole file like you asked :(

src/Components/Modules/ArenaLength.cpp
```diff
@@ -109,12 +124,7 @@ namespace Components
 		Utils::Hook(0x632799, ArenaLength::ArenaMapOffsetHook3, HOOK_JUMP).install()->quick();
 		Utils::Hook(0x4064B2, ArenaLength::ArenaMapOffsetHook4, HOOK_JUMP).install()->quick();
 
-		Utils::Hook::Set<BYTE>(0x4A95F8, 32);
-
+		Utils::Hook::Set<BYTE>(0x4A95F8, sizeof(Game::newMapArena_t::mapName));
```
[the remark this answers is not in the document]
Rackover: In the general case I would agree with you, however in that specific case this is writing the size of the string to pass it to Info_ValueForKey so imo using the sizeof makes more sense (it's used as a string of constant size, not as an array, even though the difference is purely semantic). Ill add ARRAYSIZE to please you but wrong move imo

src/Components/Modules/Zones.cpp
```diff
@@ -3624,34 +3630,34 @@ namespace Components
...
+					Utils::Hook::Call<void(int)>(0x4278A0)(false);
+				}
+			}, HOOK_CALL).install()->quick();
 
-		// Change stream for images
-		Utils::Hook(0x4D3225, [] ()
-		{
-			Game::DB_PushStreamPos((Zones::ZoneVersion >= 332) ? 3 : 0);
-		}, HOOK_CALL).install()->quick();
+			// Change stream for images
```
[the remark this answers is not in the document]
Rackover: Okay, I addressed your review and gave you write access on the branch, don't fuck it up 🙋‍♀️

diamante0018 (dismissed):

diamante0018 (approved):

PR #851, opened 2023-03-18
Remove patreon name override for now

Rackover:
We can still add it back later


diamante0018:
https://github.com/XLabsProject/iw4x-client/pull/852

PR #1038, opened 2023-05-14
[General]: Remove broken patch

diamante0018:


diamante0018:
After consulting the greatest minds available, and analysing a minidump, I come to the conclusion this patch does not solve any known exploit. The comments from V2 indicate they did not know what this function does for certain (neither do I).
What I do know is that I got so far 2 minidumps that show that around this patch a deadlock occurs.
Since this issue is IW4x exclusive and this is the only patch that comes to mind, I reached the conclusion it is the cause.
Since I can't find any way to exploit any bugs and after confirm the stability of IW4x post-removal of this hook, the benefits of removing this patch outweigh any risk posed by not removing the hook.

